PR #3, opened 2016-06-16
Fix syntax error in shed_install.py

mvdbeek:


jmchilton:
Awesome sauce, thanks @mvdbeek!

PR #4, opened 2016-06-17
Add script for generating tool_list.yml from galaxy instances

mvdbeek:
This may be useful for deployers wishing to replicate a server,
or for tracking the currently installed tool list in a VCS.


jmchilton:
Awesome sauce - thanks a bunch!

PR #5, opened 2016-07-12
generate_tool_list_from_ga_workflow_files.py

drosofff:
ping @jmchilton 
As promised here is the script that generates a yml list of tools from one or several .ga workflow files


martenson:
awesome, thank you very much @drosofff !

jmchilton:
Indeed - thanks!

PR #6, opened 2016-08-16
Add possibility to enable/disable installation of conda or toolshed p…

mvdbeek:
…ackages, split up functions.

Ping @jmchilton


jmchilton:
Awesome - thanks for continuing to enable this conda transition!

PR #7, opened 2016-08-18
Data table reload

mvdbeek:
Reload the data table after a data manager tool has run,
so that you can chain e.g downloading a reference fasta with creating the corresponding index.
Depends on https://github.com/galaxyproject/bioblend/pull/196 and a new bioblend release, I guess.


mvdbeek:
Closing this, since it is probably a limitation of galaxy's IPC, where handlers fail to communicate with a uwsgi process. It does work on instances that use e.g. rabbitmq, or the simple run.sh setup.

Issue #10, opened 2016-08-26
Fail with more meaningful message if tool doesn't exist in toolshed


mvdbeek:
Currently this will just fail with an index error in create_tool_install_payload

Issue #9, opened 2016-08-26
Use kwargs for bioblend


mvdbeek:
The latest bioblend adds one more kwarg to the install_repository_revision method.
We are passing argument values to bioblend, so this causes some unexpected behaviour (tool_panel_section_label turns into tool_panel_section_id) when using the old script in ansible-galaxy-tools. We should use kwargs anyway to have more obvious errors.

We should also consider deprecating the tool_panel_id, lot's of people on the mailing list appear to be struggling with the old role because of this.

PR #11, opened 2016-08-26
More robust shed-install script, install dependencies by default, improve logging

mvdbeek:
- Use bioblend keyword arguments, this should help prevent the problem in https://github.com/galaxyproject/ansible-galaxy-tools/pull/34

- Better checks and error messages when an invalid tool has been specified

- Flip the install_tool_dependencies switch to skip_tool_dependencies, so that by default classic toolshed packages will be used

- Add an option to skip the changeset revision retrieved from an existing galaxy instance. This list could be used to update all installed tools (Except that we specifically skip tools that are already installed ...)

This addresses #9, #10 and #13

### DIFF
--- a/ephemeris/get_tool_list_from_galaxy.py
+++ b/ephemeris/get_tool_list_from_galaxy.py
@@ -13,15 +13,17 @@ class GiToToolYaml:
     def __init__(self, url,
                  output_file,
                  include_tool_panel_section_id=False,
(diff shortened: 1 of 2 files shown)


mvdbeek:
@afgane @jmchilton I think this is ready from my side.
Could we make a new release for ephemeris soon ?

PR #14, opened 2016-08-30
tool_panel_section_name --> tool_panel_section_label

nturaga:
as per https://github.com/galaxyproject/ansible-galaxy-tools/issues/35


afgane:
I believe this line will also need to be updated: https://github.com/nturaga/ephemeris/blob/1c2a4d3b15c9e7966a1022564ef1ebdc9553b200/ephemeris/get_tool_list_from_galaxy.py#L96

mvdbeek:
Good catch, thanks @afgane @nturaga

PR #17, opened 2016-08-31
Get tool list entrypoint and improvements

mvdbeek:
- tool_shed_url will have a trailing slash added
- add https if tool_shed_url doesn't list protocol
- get-tool-list entrypoint

Ping @jmchilton


jmchilton:
Marvelous!

PR #20, opened 2016-09-25
Add a script to populate data libraries

bgruening:
I would like to ship this inside our Docker container so we can use this to get training data into training instances during build time.

### DIFF
--- a/ephemeris/library_data_example.yaml
+++ b/ephemeris/library_data_example.yaml
@@ -1,0 +1,9 @@
+libraries:
+    - name: library_name_test_1
+      files:
+        - https://zenodo.org/record/60520/files/dbSNP_138.hg19.vcf
+    - name: library_name_test_2
+      files:
+        - https://zenodo.org/record/60520/files/mother.bam
+        - https://zenodo.org/record/60520/files/father.bam
+        - https://zenodo.org/record/60520/files/patient.bam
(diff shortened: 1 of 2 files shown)


ephemeris/library_data_example.yaml
```diff
@@ -0,0 +1,9 @@
+libraries:
+    - name: library_name_test_1
+      files:
+        - https://zenodo.org/record/60520/files/dbSNP_138.hg19.vcf
+    - name: library_name_test_2
+      files:
```
mvdbeek: @bgruening can you call this urls? Just in case we extend this to (local) files at some point.

mvdbeek:
Nice 👍

ephemeris/library_data_example.yaml
bgruening: Interesting idea, but in this case we could automatically figure out the local file and do not need to distinguish between local and url, isn't it?

```
if os.path.exists():
    cp ...
else:
   wget ...
```
A folder would be an other idea, in this case I actually like the `files` and `folder` naming. What do you think?

mvdbeek: Yeah, that's true. OK then!

bgruening:
Mh, tests are failing, are they supposed to work?

mvdbeek:
Yes, these are all lint problems:
https://travis-ci.org/galaxyproject/ephemeris/jobs/162566087#L192

bgruening:
Thanks!

PR #23, opened 2016-10-22
Mention mimum required galaxy versions

### LINKED ISSUE #22: Requires Galaxy >16.04
Not directly clear from the script, but older Galaxy versions will not include the tool revision and the `generate_tool_list_from_ga_workflow_files.py` and `get-tool-list` will silently produce empty files.
see https://github.com/galaxyproject/galaxy/pull/1752 and https://github.com/galaxyproject/galaxy/pull/1636

mvdbeek:
Should close #22.


bgruening:
Thanks @mvdbeek!

PR #26, opened 2017-03-25
enable global options in a tool yaml file

bgruening:
This PR will enable global options in a tool yaml files. Means 
```
  install_repository_dependencies: true
  install_resolver_dependencies: true
  install_tool_dependencies: false
```
can be specified as top arguments next to the Galaxy instance and API key. It will also change the default setting, to install resolver-dependencies and not tool-dependencies anymore.

### DIFF
--- a/ephemeris/shed_install.py
+++ b/ephemeris/shed_install.py
@@ -44,7 +44,9 @@ from bioblend.galaxy.toolshed import ToolShedClient
 from bioblend.toolshed import ToolShedInstance
 
 MTS = 'https://toolshed.g2.bx.psu.edu/'  # Main Tool Shed
-
+INSTALL_TOOL_DEPENDENCIES = False
+INSTALL_REPOSITORY_DEPENDENCIES = False
+INSTALL_RESOLVER_DEPENDENCIES = True
 
 class ProgressConsoleHandler(logging.StreamHandler):
     """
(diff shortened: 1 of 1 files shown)


bgruening:
@mvdbeek if you have some time, can you look at this please?

afgane:
@bgruening I think this is really needed and hence glad to see it. I'd just suggest to include a link to the docs explaining those or add a one line comment about each (because for a deployer that's not very invested in the tool dev, the differences between the three options are not really clear).

bgruening:
@afgane where should this documentation go? In the example yaml?

afgane:
That would probably be best. Alternatively, above each definition of the global var? (e.g., https://github.com/galaxyproject/ephemeris/pull/26/files#diff-3e2dc1314104230489feb0f417f2d40fR47)

mvdbeek:
Maybe we should set repository_dependencies by default to true, to account for suites? Other than that looks good to me.

bgruening:
Added some docs and changed the default value. Thanks for the review!

jmchilton:
Awesome - thanks a bunch @bgruening!

afgane:
I guess I should have been more explicit because what I meant by docs was to specify differences between tool, repository, and resolver dependencies...

bgruening:
@afgane this are TS secrets no one knows exactly ;)
I will try my best to add those.

afgane:
That's how it feels every time I come cross them...

BTW, I'm really glad to see this switch that happened here.

PR #30, opened 2017-04-05
Run data managers

bgruening:
This scripts let you define some data managers with different `items` and parameters to populate Galaxy instances like Galaxy flavors.

### DIFF
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -34,7 +34,7 @@ def main(args):
     gi = GalaxyInstance(url=url, email=args.user, password=args.password)
 
     # should test valid connection
-    log.info("List of valid histories: %s" % gi.histories.get_histories())
+    log.info("List of valid histories: %s" % gi.users.get_current_user())
 
     conf = yaml.load(open(args.config))
     for dm in conf.get('data_managers'):
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -7,7 +7,7 @@ import time
 try:
     from urllib.parse import urlparse
 except ImportError:
-    from urlparse import urlparse
+    import urlparse
 
 import yaml
 from bioblend.galaxy import GalaxyInstance
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -1,0 +1,85 @@
+##!/usr/bin/env python
+
+import os
+import re
+import sys
+import time
+import argparse
+import urlparse
+import yaml
+import logging as log
+from bioblend.galaxy import GalaxyInstance
+from bioblend.galaxy.client import ConnectionError
+
+DEFAULT_URL = "http://localhost"
+
+
+def wait( gi, job ):
+    """
+        Waits until a data_manager is finished or failed.
+        It will check the state of the created datasets every 30s.
+    """
+    while True:
+        value = job['outputs']
+        # check if the output of the running job is either in 'ok' or 'error' state
+        if gi.datasets.show_dataset( value[0]['id'] )['state'] in ['ok', 'error']:
+            break
+        log.info('Data manager still running.')
+        time.sleep(30)
+
+
+def main( args ):
+    url = args.galaxy or DEFAULT_URL
+    gi = GalaxyInstance(url=url, email=args.user, password=args.password)
+
+    # should test valid connection
+    log.info("List of valid histories: %s" % gi.histories.get_histories())
+
+    conf = yaml.load(open(args.config))
+    for dm in conf.get('data_managers'):
+        for item in dm.get('items', [None]):
+            dm_id = dm['id']
+            params = dm['params']
+            log.info('Running DM: %s' % dm_id)
+            inputs = dict()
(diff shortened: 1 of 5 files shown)


bgruening:
@afgane do you also have the rights to upload a new version to pypi?

bgruening:
ping @bebatut 
This might be interesting for Aisam as well.

jmchilton:
@bgruening I think I only have rights to upload to pypi currently - if you get the HISTORY.rst file in shape I'll create a new version.

bgruening:
@jmchilton I added it here: https://github.com/galaxyproject/ephemeris/commit/980a274d401ddd35c754421fcdf918b20355bc31

jmchilton:
Thanks @bgruening - I tweaked that a bit. When this gets merged I will release this.

bgruening:
Thanks John, waiting for @afgane or @mvdbeek and then we are ready to go.

ephemeris/run_data_managers.py
```diff
@@ -0,0 +1,82 @@
...
+        log.info('Data manager still running.')
+        time.sleep(30)
+
+
+def main(args):
+    url = args.galaxy or DEFAULT_URL
+    gi = GalaxyInstance(url=url, email=args.user, password=args.password)
+
+    # should test valid connection
+    log.info("List of valid histories: %s" % gi.histories.get_histories())
```
mvdbeek: Maybe `gi.users.get_current_user()` is a little more useful here?

ephemeris/run_data_managers.py
```diff
@@ -0,0 +1,82 @@
...
+                inputs.update({key: value})
+
+            # run the DM-job
+            job = gi.tools.run_tool(history_id=None, tool_id=dm_id, tool_inputs=inputs)
+            wait(gi, job)
+            log.info('Reloading data managers table.')
+            for data_table in dm.get('data_table_reload', []):
+                # reload two times
+                for i in range(2):
+                    gi.make_get_request(urlparse.urljoin(url, 'api/tool_data/%s/reload' % data_table))
```
mvdbeek: This is only going to work if either an external message queue is running, or galaxy is running in `run.sh` mode.
Our current standard setup can't send the reload task from uwsgi to the job handlers. 
So if you for example add a new genome with a genome data manager, and then try to create a bowtie2 index for that genome there is a good chance this will fail because the handler doesn't know about the new genome.
I have added automatic reloading of changed data tables (without an explicit API call) in https://github.com/galaxyproject/galaxy/pull/3533, but right now we still need to have watchdog installed for this to work (we should make a wheel for that).

mvdbeek:
Apart from my 2 comments (the second one is definitely not blocking a merge) this looks good, thanks @bgruening.

ephemeris/run_data_managers.py
bgruening: Maybe I was lucky but this worked for me several time in the Galaxy-Docker setup and the external message queue is not properly configured.
We could add a note to the script that watchdog needs to be installed? Would this be ok?

ephemeris/run_data_managers.py
```diff
@@ -0,0 +1,82 @@
...
+                    time.sleep(5)
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Running Galaxy data managers in a defined order with defined parameters.')
+    parser.add_argument("-v", "--verbose", help="Increase output verbosity.",
+                        action="store_true")
+    parser.add_argument("--config", required=True, help="Path the YAML config file.")
```
afgane: This should probably be a bit more explicit to say _[Required] Path to the YAML config file with the list of data managers and data to install_. There's also a missing _to_ in the current help text.

Sorry to be a stickler on documentation but if a new person comes along and wants to use this, the docs is the first thing they'll see and based on that make a decision if something is usable or not.

ephemeris/run_data_managers.py
afgane: This is not going to work with the current imports. Need to just `import urlparse` vs. `from urlparse import urlparse`.

ephemeris/run_data_managers.py
bgruening: Sure, I added your sentence. It was required before so I left that as it is. argparse will give indication that this parameter is required.

ephemeris/run_data_managers.py
bgruening: You mean the python2 case? Let me test this and fix it.

bgruening: Changed

afgane: Hmm, actually, it's not going to work on py3 either because `urljoin` is not a module within `urlparse.urlparse`. It may be best to import `urljoin` directly for both py2 & 3 and then change this line to just use it:
```
try:
  from urllib.parse import urljoin
except ImportError:
  from urlparse import urljoin
```

afgane:
Thanks for all the edits @bgruening.

jmchilton:
Version 0.5.0 of ephemeris has been published - https://pypi.python.org/pypi/ephemeris.

bgruening:
@jmchilton thanks - unfortunately I think we need https://github.com/galaxyproject/ephemeris/pull/32 as well :(

PR #32, opened 2017-04-06
Clean up the main() function

bgruening:
It seems that the current setup.py setup needs to have a clean main() function.
Anyone know what is the preferred Python way to handle such a setup? I have seen many projects that writes the `bin/*` on there own.

### DIFF
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -60,6 +60,7 @@ def run_dm(args):
                     gi.make_get_request(urljoin(url, 'api/tool_data/%s/reload' % data_table))
                     time.sleep(5)
 
+
 def main():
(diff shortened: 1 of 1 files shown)


jmchilton:
@bgruening For what it is worth - I prefer this version to previous. I'm +1 on this as is but if you want to spend more time coming up with a best practice Python standard let me know.

bgruening:
@jmchilton I'm fine with this. It was a little bit unexpected to have a main() function without parameter, I was just wondering if this is the best-practice or if there is anything else.
Please feel free to merge and release a new version :)

bgruening:
Thanks!!!

jmchilton:
https://pypi.python.org/pypi/ephemeris

PR #36, opened 2017-04-17
Add initial test against the Galaxy Docker image

bgruening:
This add simple tests to all tools. Things can be improved, but we should have a nice start here.
Also I fixed the python3 combat here, which was not really working before.

I was hitting one strange Travis bug and need to add a workaround.
New version would be great! :)


jmchilton:
Magnificent - I'll do a new release right away.

bgruening:
@jmchilton thanks John this was quick.
This will now enable every trainer very easy production of training flavors. One example here https://github.com/bgruening/galaxy-rna-workbench/blob/master/Dockerfile, tours, workflows and test data integrated.

PR #37, opened 2017-04-29
Add sleep script

erasche:
Lots of the flavour images do a `startup_lite && sleep 30 && ...` in order to do things on a live database. This number can be quite variable. 30s was fine for me, but wasn't for a colleague who had to increase it.

This script can be used in place of `sleep 30` and will sleep until `/api/version` is available. This *should* be much more reliable.

A timeout option is provided to prevent abusing build resources if something else is at fault (e.g. a missing package allowed startup_lite to succeed initially but fail silently later).

### DIFF
--- a/ephemeris/sleep.py
+++ b/ephemeris/sleep.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+import requests
+import sys
+import time
+
+from argparse import ArgumentParser
+
+from .common_parser import get_common_args
+from .shed_install import setup_global_logger, _disable_external_library_logging
+
+
+def _parse_cli_options():
+    """
+    Parse command line options, returning `parse_args` from `ArgumentParser`.
+    """
+    parent = get_common_args(login_required=False)
+    parser = ArgumentParser(parents=[parent], usage="usage: python %(prog)s <options>")
+    parser.add_argument("--timeout",
+                        default=0, type=int,
+                        help="Galaxy startup timeout. The default value of 0 waits forever")
+    return parser.parse_args()
+
+
+def main():
+    global log
+    _disable_external_library_logging()
+    log = setup_global_logger(include_file=True)
+    options = _parse_cli_options()
+
+    count = 0
+    while True:
+        try:
+            result = requests.get(options.galaxy + '/api/version').json()
+            if options.verbose:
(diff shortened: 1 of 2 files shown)


ephemeris/sleep.py
```diff
@@ -0,0 +1,53 @@
...
+    parser.add_argument("--timeout",
+                        default=0, type=int,
+                        help="Galaxy startup timeout. The default value of 0 waits forever")
+    return parser.parse_args()
+
+
+def main():
+    global log
+    _disable_external_library_logging()
+    log = setup_global_logger(include_file=True)
```
bgruening: Is this logging needed here? It is not used.

ephemeris/sleep.py
```diff
@@ -0,0 +1,53 @@
...
+
+def _parse_cli_options():
+    """
+    Parse command line options, returning `parse_args` from `ArgumentParser`.
+    """
+    parent = get_common_args(login_required=False)
+    parser = ArgumentParser(parents=[parent], usage="usage: python %(prog)s <options>")
+    parser.add_argument("--timeout",
+                        default=0, type=int,
+                        help="Galaxy startup timeout. The default value of 0 waits forever")
```
bgruening: Mention that it is in seconds?

ephemeris/sleep.py
```diff
@@ -0,0 +1,53 @@
...
+from .common_parser import get_common_args
+from .shed_install import setup_global_logger, _disable_external_library_logging
+
+
+def _parse_cli_options():
+    """
+    Parse command line options, returning `parse_args` from `ArgumentParser`.
+    """
+    parent = get_common_args(login_required=False)
+    parser = ArgumentParser(parents=[parent], usage="usage: python %(prog)s <options>")
```
bgruening: Can we add here a small description about what this tool is doing or make the tool name a little bit more descriptive?

bgruening:
Nice idea!!!

ephemeris/sleep.py
erasche: Planned to use it for output and then forgot... will fix.

bgruening:
Thanks @erasche I think we should rename the script as well, but we can do this later!

PR #38, opened 2017-05-01
install_resolver_dependencies keys are printed in the tool_list.yml

drosofff:
with a nasty trick to preserve a more readable order of keys in the yml file (name first)
install_resolver_dependencies is always set to True

### DIFF
--- a/ephemeris/generate_tool_list_from_ga_workflow_files.py
+++ b/ephemeris/generate_tool_list_from_ga_workflow_files.py
@@ -64,7 +64,8 @@ def print_yaml_tool_list(tool_dictionary, output_file):
         F.write(yaml.safe_dump(tool_dictionary, default_flow_style=False))
     return
 
-def reduce_tool_list (tool_list):
+
+def reduce_tool_list(tool_list):
     for current_tool in tool_list:
         for tool in tool_list:
             if current_tool is tool:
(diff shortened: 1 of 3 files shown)


ephemeris/generate_tool_list_from_ga_workflow_files.py
```diff
@@ -49,14 +48,24 @@ def translate_workflow_dictionary_to_tool_list(tool_dictionary, panel_label):
     tool_list = []
     for tool in starting_tool_list:
         sub_dic = {'name': tool['name'], 'owner': tool['owner'], 'revision': tool['changeset_revision'],
-                   'tool_panel_section_label': panel_label, 'tool_shed_url': 'https://'+tool['tool_shed']}
+                  'tool_panel_section_label': panel_label, 'tool_shed_url': 'https://'+tool['tool_shed'],
+                  'zinstall_resolver_dependencies': 'True'}
```
bgruening: the `z` looks strange :)

drosofff: Yes this is the nasty trick I mentionned... This is not a bug, this is a feature :-/

bgruening:
@drosofff do you mind to add a test for this tool?

ephemeris/generate_tool_list_from_ga_workflow_files.py
drosofff: see line 63

drosofff:
@bgruening No I don't. I put it in my todo list !

drosofff:
@bgruening here is a test. Works on my fork, should work on the upstream.

bgruening:
@drosofff thanks for the tests it looks good.
I have a general comment/hint that hopefully makes this patch easier. Have you seen this patch? https://github.com/galaxyproject/ephemeris/pull/26

With this in place you only need to specify it once on top of all tools and don't need to mess with he ordering :)

bgruening:
Thanks @drosofff!

PR #41, opened 2017-05-18
Enable api key on run_data_managers

rhpvorderman:
run_data_managers has an -a or --api-key flag inherited from common args but did not allow for using the api_key.

### DIFF
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -33,12 +33,15 @@ def wait(gi, job):
 
 def run_dm(args):
     url = args.galaxy or DEFAULT_URL
-    if args.api_key != "":
+    if args.api_key:
         gi = GalaxyInstance(url=url, key=args.api_key)
     else:
         gi = GalaxyInstance(url=url, email=args.user, password=args.password)
-        # should test valid connection
-        log.info("List of valid histories: %s" % gi.users.get_current_user())
+    
+    # should test valid connection
+    # The following should throw a ConnectionError when invalid API key or password
+    genomes = gi.genomes.get_genomes()
+      
 
     conf = yaml.load(open(args.config))
     for dm in conf.get('data_managers'):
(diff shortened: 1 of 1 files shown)


ephemeris/run_data_managers.py
```diff
@@ -33,10 +33,12 @@ def wait(gi, job):
...
     url = args.galaxy or DEFAULT_URL
-    gi = GalaxyInstance(url=url, email=args.user, password=args.password)
-
-    # should test valid connection
-    log.info("List of valid histories: %s" % gi.users.get_current_user())
+    if args.api_key != "":
+        gi = GalaxyInstance(url=url, key=args.api_key)
+    else:
+        gi = GalaxyInstance(url=url, email=args.user, password=args.password)
+        # should test valid connection
```
afgane: This should probably be moved outside the `if/else`.

ephemeris/run_data_managers.py
```diff
@@ -33,10 +33,12 @@ def wait(gi, job):
 
 def run_dm(args):
     url = args.galaxy or DEFAULT_URL
-    gi = GalaxyInstance(url=url, email=args.user, password=args.password)
-
-    # should test valid connection
-    log.info("List of valid histories: %s" % gi.users.get_current_user())
+    if args.api_key != "":
```
bgruening: This can be `if args.api_key:`.

Thanks @rhpvorderman!

ephemeris/run_data_managers.py
rhpvorderman: Yes. But the test is specific for the current user. When using the master api key: there is no current user. 
Currently the master api key cannot run the data managers because it does not have admin rights. But I hope it will get special permission to run the data managers as well in the future.

bgruening:
@rhpvorderman looks good to me! Can you please fix the pep8 linting.

afgane:
Thanks for all the followups!

Issue #1, opened 2016-06-15
Improve Project Structure


jmchilton:
- [x] Implement a README.
- [x] Review and fix all Travis tests.
- [x] Generate and build documentation.
- [x] Setup a readthedocs site for this.
- [x] Write a couple tests.
- [x] Enable coverage testing, register with coveralls.

jmchilton:
@bgruening took a huge step forward on this with #36. If we find a way to just dump the script help text to RST for readthedocs I think we are very close to having a decent, well-tested user-facing application here.

rhpvorderman:
I'm working on a project that depends on tools in ephemeris. I am willing to look at the documentation if I am not duplicating effort and someone gives me directions as to where the readthedocs documentation is/should be stored.

bgruening:
@rhpvorderman any documentation is welcome at this point and I don't think anyone is working on this yet. 
Thanks!

jmchilton:
I'll set up a Sphinx outline and create a readthedocs today to enable contributions of documentation - thanks @rhpvorderman!

jmchilton:
@rhpvorderman I have opened a PR for some initial Sphinx docs here (https://github.com/galaxyproject/ephemeris/pull/44). I think obvious wins are docstrings and improved command help - I'm pretty sure both of these kinds of documentation would be very valuable and both will ultimately land up on Sphinx or in users hands directly. Thanks so much for contributing - I'm looking forward to seeing what you are building!

rhpvorderman:
> If we find a way to just dump the script help text to RST for readthedocs I think we are very close to having a decent, well-tested user-facing application here.

This is enabled in #46 . Using sphinx-argparse to write shed-install and run-data-manager's help texts to rst. It should be easily portable to other functions as well.

rhpvorderman:
Coverage is now tested and reported. Some tests need to be converted to the new pytest format. But all in all. The project now has the desired structure.

PR #44, opened 2017-06-09
Setup Sphinx docs so this can be published to RTD.

jmchilton:
This is based on the Pulsar/Planemo template so this fills in some initial content - install directions, code of conduct, contributing description, docs for process for publishing to PyPI, and establishing a governance policy (@bgruening as "Benevolent Dictator for Now" with a small "committer" group (@afgane, @drosofff, @jmchilton, @bgruening, @mvdbeek) - everyone who has added more than 100 lines of code).

I think another obvious win for documentation would be dump the ``--help`` for all the commands to the docs. I do this with Planemo here (http://planemo.readthedocs.io/en/latest/commands.html) and it works well I think. I don't know if the lack of Click here would make that more difficult now.

xref #1


erasche:
I'm happy to write a short tutorial on using the sleep command / example uses / etc. Have been waiting for this to be solved :) could really benefit from some how to / cookbook type things.

rhpvorderman:
If this request was pulled I would gladly contribute with documentation on the usage of shed-install and run-data-managers.

jmchilton:
Thanks for the offers @erasche and @rhpvorderman - sorry for not being better about documenting these things from the get go. It is hard to know what ideas are going to catch and what ideas are going to fail - so I tend not to document things well unless I know it will be read. I will try to configure RTDs now.

PR #49, opened 2017-08-18
Use https by default

bgruening:


rhpvorderman:
This will not work on galaxies that do not have https. Also nginx should route any http requests to https when the galaxy is configured to use https.

erasche:
@rhpvorderman it does work for galaxies without https, they just have to specify `http://` when they're referencing their galaxy. Nice to encourage security by default, maybe?

rhpvorderman:
> Routing of http to https only works for GETs, and POST requests will not be re-executed.

@erasche Then https galaxies do not work well in the current situation. And this pull request flips the current situation. Instead of having to specify `https://` to reference your galaxy you don't have to that anymore. 

> Nice to encourage security by default, maybe?

Yes, indeed.

erasche:
yeah, if I'm being honest, I'd personally prefer that protocol-less URLs are just rejected always since they're somewhat meaningless / ambiguous.

rhpvorderman:
Something like this?
```python
    if not galaxy_url.startswith('http'):
        raise ValueError('URL should start with http:// or https://')
```

rhpvorderman:
Or maybe
```python
    if not galaxy_url.startswith('http'):
        log.warning('URL should start with http:// or https://. https:// chosen by default.')
        galaxy_url = 'https://' + galaxy_url
```

erasche:
:+1:

afgane:
I'm in favor of @rhpvorderman's suggestion if this is to become the default.

bgruening:
Done!

PR #51, opened 2017-09-05
Run-data-managers now checks whether an entry in the yaml is already installed.

### LINKED ISSUE #50: Data manager does not check if index is already populated.
The shed-install command works really nice: you feed it a yaml and it installs the tools. Oops! You forgot a tool! No worries, just add it to the yaml and run shed-install again. It will skip the already installed tools.

This behaviour is not reproduced in the run-data-managers tool. If you run a yaml again, it will start indexing all the genomes in there. Even if they are already indexed. This is annoying if you have human genomes on the list which take a long time to index.

rhpvorderman:
Fixes #50

### DIFF
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -20,6 +20,7 @@ except ImportError:
 
 import yaml
 from bioblend.galaxy import GalaxyInstance
+from bioblend.galaxy.tool_data import ToolDataClient
 
 from .common_parser import get_common_args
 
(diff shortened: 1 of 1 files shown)


bgruening:
@rhpvorderman wonderful! Can we enhance the tests so that a second install of a DM does nothing?

rhpvorderman:
I could check. I see that there are now no tests at all for run_data_managers. I think I will use the fetch genome data manager to get some virus or something and do it twice.
Maybe pull #52 first so the test sample can be a bit more advanced?

rhpvorderman:
Added tests with advanced templating in #53

PR #52, opened 2017-09-05
Allow advanced jinja templating

rhpvorderman:
Swapped out the regex substitution for jinja2 templating.
This allows for more functional templating as shown in `tests/run_data_managers.yaml.sample`.
It is still compatible with the old format.
This is in the first two commits.

Since having more than 3 genomes already gives some intensive copy and paste clutter also a `genomes` variable is added. see `tests/run_data_managers.yaml.sample.advanced`.
This was added in the last two commits.

This will have merge conflicts with #51 , but I already fixed that in my own fork of ephemeris (in the `fixes` branch). So I can update as soon as #51 is merged.


bgruening:
Cool beans! Let's see if all tests are passing.

rhpvorderman:
#53

PR #53, opened 2017-09-11
Advanced yaml templating and checking of already installed genomes.

rhpvorderman:
Combination of #51 and #52  with tests.


ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+        if field[column_index] == entry:
+            return True
+    return False
+
+
+def get_name_from_inputs(input_dict):
+    '''Returns the value that will most likely be recorded in the "name" column of the datatable. Or returns None'''
+    possible_keys = ['name', 'sequence_name']  # In order of importance!
+    for key in possible_keys:
+        if bool(input_dict.get(key)):
```
bgruening: `if input_dict.get(key, False)`

bgruening: Or `if key in input_dict`

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+            return input_dict.get(key)
+    return None
+
+
+def get_value_from_inputs(input_dict):
+    '''Returns the value that will most likely be recorded in the "value" column of the datatable. Or returns None'''
+    possible_keys = ['value', 'sequence_id', 'dbkey']  # In order of importance!
+    for key in possible_keys:
+        if bool(input_dict.get(key)):
```
bgruening: if key in input_dict

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+    return False
+
+
+def get_name_from_inputs(input_dict):
+    '''Returns the value that will most likely be recorded in the "name" column of the datatable. Or returns None'''
+    possible_keys = ['name', 'sequence_name']  # In order of importance!
+    for key in possible_keys:
+        if bool(input_dict.get(key)):
+            return input_dict.get(key)
+    return None
```
bgruening: return False?

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+
+
+def get_value_from_inputs(input_dict):
+    '''Returns the value that will most likely be recorded in the "value" column of the datatable. Or returns None'''
+    possible_keys = ['value', 'sequence_id', 'dbkey']  # In order of importance!
+    for key in possible_keys:
+        if bool(input_dict.get(key)):
+            return input_dict.get(key)
+    return None
```
bgruening: return False

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+
+
+def input_entries_exist_in_data_tables(tool_data_client, data_tables, input_dict):
+    '''Checks whether name and value entries from the input are already present in the data tables.
+    If an entry is missing in of the tables, this function returns False'''
+    value_entry = get_value_from_inputs(input_dict)
+    name_entry = get_name_from_inputs(input_dict)
+
+    # Return False if name and value entries are both none
+    if not bool(value_entry) and not bool(name_entry):
```
bgruening: remove bool()

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+        return False
+
+    # Check every data table for existance of name and value
+    # Return False as soon as entry is not present
+    for data_table in data_tables:
+        if bool(value_entry):
```
bgruening: remove bool

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,76 @@ def wait(gi, job):
...
+    # If all checks are passed the entries are present in the database tables.
+    return True
+
+
+def parse_items(items, genomes):
+    if bool(genomes):
+        items_template = Template(json.dumps(items))
+        rendered_items = items_template.render(genomes=json.dumps(genomes))
+        # Remove trailing " if present
+        if rendered_items.startswith('"') and rendered_items.endswith('"'):
```
bgruening: strip('"')

ephemeris/run_data_managers.py
```diff
@@ -48,32 +128,42 @@ def run_dm(args):
...
-            log.info('Reloading data managers table.')
-            for data_table in dm.get('data_table_reload', []):
-                # reload two times
-                for i in range(2):
-                    gi.make_get_request(urljoin(gi.url, 'api/tool_data/%s/reload' % data_table))
-                    time.sleep(5)
+            data_tables = dm.get('data_table_reload', [])
+            # Only run if not run before.
+            if input_entries_exist_in_data_tables(tool_data_client, data_tables, inputs) and not args.overwrite:
+                log.info('%s already run for %s' % (dm_id, str(inputs)))
```
bgruening: `is already available`...

str() cast is not needed if you use %s

ephemeris/run_data_managers.py
```diff
@@ -48,32 +128,42 @@ def run_dm(args):
...
+            else:
+                log.info('Running DM: "%s" with parameters: %s' % (dm_id, str(inputs)))
```
bgruening: str() not needed

rhpvorderman:
This is strange. First test job finishes fine. Other jobs crash?

mvdbeek:
> This is strange. First test job finishes fine. Other jobs crash?

If the test is using a setup with separate handlers the data table reload won't work. You'll have to activate watch_tool_data_dir the galaxy.ini https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.ini.sample#L326

rhpvorderman:
Thanks @mvdbeek . Let's see if it works.

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
     - git config --global user.name "Travis Test User"
     - git config --global user.email "galaxy_test@galaxyproject.org"
     - pip install tox coveralls
-    - docker run -d -p 8080:80 bgruening/galaxy-stable
+    - docker run -d -p 8080:80 bgruening/galaxy-stable -e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True
```
mvdbeek: Doesn't `-e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True` need to be before the image ?

rhpvorderman: My bad!

mvdbeek:
Hmm, does docker-galaxy-stable include watchdog ? That may be required ...

rhpvorderman:
It does. Because ansible-galaxy-extras installs this by default.
The strange thing is that the run-data-managers command can reload the data tables itself:
`DEBUG:urllib3.connectionpool:http://localhost:8080 "GET /api/tool_data/fasta_indexes/reload?key=admin HTTP/1.1" 200 None`
However. This does not have a 100% success rate on the docker container. 
Some test jobs fail, while others succeed. It seems to be random.

mvdbeek:
> However. This does not have a 100% success rate on the docker container.
> Some test jobs fail, while others succeed. It seems to be random.

Last time I checked this will only work for the current web request, which in turn won't be able to communicate the change to the job handlers, so it isn't strange, just unfortunate.

rhpvorderman:
Is there a way to fix this with settings so the test can be run?

mvdbeek:
I'm checking this out now.

mvdbeek:
So indeed watchdog is properly installed and working, but when we install a new data manager we add a new data table, that is not being watched by watchdog. I think restarting galaxy after the data manager installation would fix this test, while I think of a way to fix this in galaxy.

mvdbeek:
I think https://github.com/rhpvorderman/ephemeris/pull/1 should work, but I haven't tested it.

rhpvorderman:
@mvdbeek Looks good! Thanks!
Like I said, ephemeris seems to work perfectly fine outside of travis. Due to limited resources the data table reloads maybe too slow. Another way to fix it would be to introduce  custom wait times for the reloading which can be changed by a command line flag. But this is a bit hackish of a solution for only one use case.

rhpvorderman:
Darn! Still some failing issues, but these seem not to be related at all to the run-data-managers program. It is toxenv.sh failing for python 3.5

mvdbeek:
Let me take a bit of the guesswork out here:
The data table reload takes milliseconds, that is not a problem. 
The issue (AFAICT) is 3 fold: 
  - If you're running galaxy using the default config in run.sh you're just using a single shared web and job handler process, so no cross-process communication is necessary to notify all processes that new data table entries are available. Installing tools / data managers and running them is not a problem. The manual reloading through the API is not necessary.
  - In a production setup with multiple processes the process that received the job (ie. running the data manager) will get the new table -- if you're just using 1 process that will be fine, but most setups use multiple processes and threads, these need to be notified somehow. We can do this 2 ways 
    -  each process can listen to changes to the data tables, or we can use an external messaging queue like rabbitmq to pass messages. @bgruening's image comes with rabbitmq, but it isn't setup properly, and maintaining rabbitmq is another additional complication, so that's not ideal.
    - we use watchdog to monitor the individual tables in each process, which means watchdog must be installed and `watch_tool_data_dir` must be set. This works fine (I've been using this in production ever since I added this option), but it only listens to tables that are known during server startup. So if you install a new data manager, but you don't restart galaxy watchdog will not listen to the new table.

For both of the latter option it appears as if hitting the data table endpoint works if you just do it often enough, because you stochastically hit the different processes/threads, but the handlers will never be updated, so when you run a job that relies on a dbkey and a fasta item, the job will simply fail, which is what we've been observing before https://github.com/galaxyproject/ephemeris/pull/53/commits/ed6a30883d89cde7c06cad7a763d9d1dbc29f111.
To not give people reading the code the wrong idea, I'd be strongly in favor of removing `tool_data_client.reload_data_table(str(data_table))`

rhpvorderman:
> we use watchdog to monitor the individual tables in each process, which means watchdog must be installed and watch_tool_data_dir must be set. This works fine (I've been using this in production ever since I added this option), but it only listens to tables that are known during server startup. So if you install a new data manager, but you don't restart galaxy watchdog will not listen to the new table.

That makes sense. That will solve a weird issue I had in my ansible provisioning script. Thanks. Will remove it from run-data-managers and shall add this bit of information to the tests.

mvdbeek:
The failing tests seem to be related to python-3.5 not being in travis' environment. Can you add  `- "3.5"` after line 6 in the .travis.yml file ?

mvdbeek:
Can you try this and remove the separate python and env section ?
```
matrix:
  include:
    - python: 2.7
      env: TOX_ENV=py27-lint
    - python: 2.7
      env: TOX_ENV=py27-lint-readme
    - python: 2.7
      env: TOX_ENV=py27
    - python: 3.5
      env: TOX_ENV=py35-lint
    - python: 3.5
      env: TOX_ENV=py35
   global:
     # do not load /etc/boto.cfg with Python 3 incompatible plugin
     # https://github.com/travis-ci/travis-ci/issues/5246#issuecomment-166460882
     - BOTO_CONFIG=/doesnotexist
```

rhpvorderman:
Now it is shed-install choking on the python3 tests...

mvdbeek:
Cool, at least we're getting somewhere. This looks like a bioblend issue.

rhpvorderman:
This piece in shed_install.py is causing the error 
```python
    tsc = tool_shed_client(gi)
    installed_revisions_list = []
    itl = tsc.get_repositories()
```
If gi is defined using bioblend 0.9.0 and so is tsc then the problem is in 0.9.0
I will check a few versions to see what works.

mvdbeek:
This is the latest version, I can fix this in bioblend.

mvdbeek:
Alright, fix is here: https://github.com/galaxyproject/bioblend/pull/251. In the meantime is there a chance we can use an api-key instead of username/password ? This should work.

ephemeris/run_data_managers.py
```diff
@@ -40,6 +50,75 @@ def wait(gi, job):
         time.sleep(30)
 
 
+def data_table_entry_exists(tool_data_client, data_table_name, entry, column='value'):
+    '''Checks whether an entry exists in the a specified column in the data_table.'''
+    try:
+        data_table_content = tool_data_client.show_data_table(data_table_name)
+    except:
```
mvdbeek: it's always preferable to use `except Exception`, or even better the exact exception class that you are expecting in this case, so that when troubleshooting we are not mislead by `'Table "%s" does not exist'`. A bunch of reasons for this is listed in this thread https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements

ephemeris/run_data_managers.py
```diff
@@ -48,32 +127,38 @@ def run_dm(args):
...
+            # Only run if not run before.
+            if input_entries_exist_in_data_tables(tool_data_client, data_tables, inputs) and not args.overwrite:
+                log.info('%s already run for %s' % (dm_id, inputs))
+            else:
+                log.info('Running DM: "%s" with parameters: %s' % (dm_id, inputs))
+                # run the DM-job
+                job = gi.tools.run_tool(history_id=None, tool_id=dm_id, tool_inputs=inputs)
+                wait(gi, job)
+                log.info('Wait 5 seconds so data tables can be reloaded.')
+                time.sleep(5)
```
mvdbeek: This isn't necessary, the integration test I've written for galaxy doesn't need a timeout at all, but if you want to be on the safe side you can set this to 1 second.

test.sh
```diff
@@ -10,16 +10,33 @@ get-tool-list --help
...
 echo "Check tool installation from workflow"
 shed-install -t result_workflow_to_tools.yaml -a admin -g http://localhost:8080
-shed-install -t result_workflow_to_tools.yaml --user admin@galaxy.org -p admin -g http://localhost:8080
+#shed-install -t result_workflow_to_tools.yaml --user admin@galaxy.org -p admin -g http://localhost:8080
+echo "Check installation of reference genomes"
+#run-data-managers --user admin@galaxy.org -p admin -g http://localhost:8080 --config tests/run_data_managers.yaml.test -v
+run-data-managers -a admin -g http://localhost:8080 --config tests/run_data_managers.yaml.test -v
+echo "Check if installation is skipped when reference genomes are already installed."
+echo "Wait for 30 seconds to allow galaxy to update its table status"
+sleep 30
```
mvdbeek: That shouldn't be necessary

mvdbeek:
@rhpvorderman have a look over the last changes, with your 👍 I'm happy to merge this (even though restarting galaxy is a little brittle it seems).
Ultimately https://github.com/galaxyproject/galaxy/pull/4617 should do away with the need for restarting.

rhpvorderman:
Wow! Great!
Thanks for all the contributions. The pull request at galaxy main also looks really nice. 
Let's merge this! Maybe I will get to #54 and as soon as that is fixed I think that warrants a new pypi package update.

rhpvorderman:
Oh wait. There are some things...

tests/test.sh
```diff
@@ -0,0 +1,49 @@
...
+TEST_DATA=${EPHEMERIS_TEST_DATA:-"$CURRENT_DIR"}
+
+run-data-managers --help
+shed-install --help
+workflow-install --help
+setup-data-libraries --help
+get-tool-list --help
+
+echo "Starting galaxy docker container"
+CID=`docker run -d -e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True -p 8080:80 bgruening/galaxy-stable`
```
rhpvorderman: This depends on the docker run executable to output the container name. This may change upstream. Therefore I prefer to use the --name flag.
Using a ENV variable for this is a good idea indeed.

tests/test.sh
```diff
@@ -0,0 +1,49 @@
...
+# Check if already installed was thrown
+echo $(cat data_manager_output.txt | grep -i "already run for")
+data_manager_already_installed=$(cat data_manager_output.txt | grep -i "already run for" -c)
+if [ $data_manager_already_installed -ne 2 ]
+    then
+        echo "ERROR: Not all already installed genomes were skipped"
+        exit 1
+fi
+# Remove running container
+docker stop $CID && docker rm $CID
```
rhpvorderman: docker rm has a force option -f which does the same.

tests/test.sh
mvdbeek: They are using this in their own documentation, and this is such a common part of the API that it would be really bad if they changed it. If you specify a name you're guaranteed to clash if the test has been running / if a user has a container of the same name running.

rhpvorderman: You are right. Will change that back.

tests/test.sh
mvdbeek: It's not entirely the same, if there is a problem stopping the container, the container won't be deleted. That may be nice for debugging (though I anticipate that's never going to happen).

tests/test.sh
mvdbeek: Similarly we shouldn't expose any ports, and get the internal container's ip address to be more robust in that regard. Certainly 8080 is not a great idea.

rhpvorderman: This can be fixed by only exposing it to localhost.

mvdbeek: That won't work on OSX. I'd prefer the `-P` option and the using inspect to grab the webport.
`docker inspect --format='{{(index (index .NetworkSettings.Ports "80/tcp") 0).HostPort}}' $CID` will give you the random port that been mapped to 80.

rhpvorderman: Even better is to use the containers internal ip address to communicate with it. Prevents port collision all together.

mvdbeek: No, that doesn't work on OSX https://docs.docker.com/docker-for-mac/networking/#i-cannot-ping-my-containers

rhpvorderman: Okay. I do not have access to a mac. so there is no way I can test whatever I write does work on OSX. Would you be so kind to implement this?

tests/test.sh
```diff
@@ -12,31 +14,36 @@ setup-data-libraries --help
 get-tool-list --help
 
 echo "Starting galaxy docker container"
-CID=`docker run -d -e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True -p 8080:80 bgruening/galaxy-stable`
+# We start the image with the -P flag that published all exposed container ports
+# to random free ports on the host, since on OS X the container can't be reached
+# through the internal network (https://docs.docker.com/docker-for-mac/networking/#i-cannot-ping-my-containers)
+CID=`docker run -d -e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True -P bgruening/galaxy-stable`
```
rhpvorderman: I guess the disadvantage is that now all docker container ports are exposed instead of only the web port. Does that really matter though, in terms of security? At least port collision is not a concern any more.

rhpvorderman:
I have a suggestion. This is an entirely different issue. It has nothing to do with run-data-managers and its functionality. 
Let's move the docker port exposure issue to another pull request and not try to fix it here.

tests/test.sh
mvdbeek: It isn't more than it was before, given that the web port is reachable with a predictable api key / username/password combination, but that's certainly something we should change.

mvdbeek:
Why not fix it here? It is only needed for this test / it works now. The security implications are not that severe, given that docker doesn't run as privileged user.

rhpvorderman:
I incorrectly assumed that this would take a long time to fix. My apologies for my impatience. But yes, since everything works now, why not fix it here?
Shall we merge this request?

mvdbeek:
Alright, this is some very nice functionality, thanks a lot.

rhpvorderman:
> Alright, this is some very nice functionality, thanks a lot.
And thank you for all the contributions to the testing of it. I would not have been able to do this by myself.

PR #55, opened 2017-09-18
Parallelization of run-data-managers. Add galaxy-wait to tests. Fix galaxy-wait issue with nginx.

### LINKED ISSUE #54: Run-data-managers only runs one job at a time.
Not very useful if installing 18 genomes at once with 5 data managers. Waiting on 90 jobs one at a time.
Since data managers may be interdependent it is maybe best to run these jobs per data manager.
So starting 18 jobs in parallel for each data manager and then wait until all jobs are finished to reload the data table and start the next data manager for the 18 genomes.

rhpvorderman:
fixes #54 

~~The stdout of run-data-managers is not up to scratch though. It does not give any output. If the -v flag is used, it gives too much output (also debug messages). Can somebody show me how to get run-data-managers too display all the info messages in stdout?~~

### DIFF
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -43,7 +43,7 @@ def wait(gi, job_list):
     """
     # Empty list returns false
     while bool(job_list):
-        finished_jobs=[]
+        finished_jobs = []
         for job in job_list:
             value = job['outputs']
             job_id = job['outputs'][0]['hid']
@@ -54,9 +54,10 @@ def wait(gi, job_list):
                 finished_jobs.append(job)
             else:
                 log.debug('Job %i still running.' % job_id)
-        # only sleep if job_list is not empty yet.
+        # Remove finished jobs from job_list.
         for finished_job in finished_jobs:
-            job_list.remove(finished_job) # Empties the list.
+            job_list.remove(finished_job)
+        # only sleep if job_list is not empty yet.
         if bool(job_list):
             time.sleep(30)
 
--- a/ephemeris/run_data_managers.py
+++ b/ephemeris/run_data_managers.py
@@ -167,7 +167,6 @@ def run_dm(args):
             if input_entries_exist_in_data_tables(tool_data_client, data_tables, inputs) and not args.overwrite:
                 log.info('%s already run for %s' % (dm_id, inputs))
             else:
-                log.info('Running DM: "%s" with parameters: %s' % (dm_id, inputs))
                 # run the DM-job
                 job = gi.tools.run_tool(history_id=None, tool_id=dm_id, tool_inputs=inputs)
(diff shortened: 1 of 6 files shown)


rhpvorderman:
Found the line to enable more output.

rhpvorderman:
Moved the "data manager still running" to debug as it clutters output. 
Removing items from a list while iterating over a list was an ugly mistake I made which caused skipping over every other job. This is now corrected.
Everything should be in order now.

rhpvorderman:
Test failed because of random error in the docker container.
Updated documentation to reflect changes made to the installer and to trigger new testing.

rhpvorderman:
Because I was impatient I already used this change in production and it works. We used it to run 10 genomes on 7 data managers. It was quite satisfying to see these jobs pushed to galaxy ten at a time instead of one by one.

Can somebody review this? Thanks in advance!

rhpvorderman:
I added some more verbose information and more error checking.
Testing on travis seems to yield inconsistent results. It does not seem to be related to the run-data-managers function. Could somebody help me with this @mvdbeek perhaps?

rhpvorderman:
Did several things:
* Tie in `galaxy-wait` in the testing. This also removes some `sleep` commands which should speed up testing and abolish unnecessary waiting.
* fix errors in `galaxy-wait` (`sleep.py`) that resulted from the docker container returning a nginx forbidden error.
* Update bioblend. In #53 the tests were changed because of an error in bioblend.  This was fixed on https://github.com/galaxyproject/bioblend/pull/251. With the new version tests can be run both on api key and user/password again on python3.
* added a small wait step to give the data tables a chance to reload between two run-data-manager commands. This should happen near-instant, but since the two `run-data-managers` commands were also run near-instantly after each other this seemed not to work on travis. A very conservative wait of 15s was added in between.

rhpvorderman:
These random errors that are due to the galaxy in docker not updating fast enough in travis are somewhat frustrating.

rhpvorderman:
Retry attempts did not result in success. The cause of the error is that while the viruses are downloaded and put in the data table, some virus do not show up in the data table. I think this is caused by the viruses being added simultaneously to the data table by different handlers. This might cause the attempt not to be registered. I tried to run the test container with only handler process to check if this fixes the problem.

rhpvorderman:
> An error occurred while generating the build script.
?

bgruening:
I think travis has some problems, I see this elsewhere as well :(

bgruening:
I will restart the test a few times in the next hours and see whats going on. There is also a new Galaxy Docker release, which might help or break things ... let's see. Eitherway - thanks a lot for this work, looks great!

bgruening:
1/6 or so failed. This is still annoying but it s not that bad. Maybe its better with the new release?

Eitherway, thanks a lot!

PR #58, opened 2017-10-26
 Add test that demonstrates that tools can be updated

mvdbeek:
by specifying a newer changeset revision.


bgruening:
Thanks @mvdbeek!

Issue #63, opened 2017-11-23
Shed-install should be able to update all the tools in a galaxy instance


rhpvorderman:
Would be nice for a cron job to keep all the tools up to date.
Changes could be made also fixing #62. 

Please let me know if this functionality is useful and I will fix this issue and #62.

I can't find a button in galaxy that just says: "update all  tools"

afgane:
I'd love to see this added.

erasche:
:+1:

Issue #61, opened 2017-11-07
0.7.1 Release


rhpvorderman:
Run-data-managers has improved significantly, an error in galaxy-wait has been fixed an unbounded variable has been fixed. ( #53 #55 #57 )
@jmchilton could you release 0.7.1 so that these changes are made available to the public?
Also I believe we have fixed a lot of issues in #1 now. Maybe the issues in #31 can now also be addressed?
Thanks!

bgruening:
@jmchilton I also got a few other PRs in. So if you can roll out a new version that would be fantastic.

rhpvorderman:
Maybe we can also get #64 in.
Also, this release will break the `shed-install` command. It will use the `shed-tools install` command instead.
By the rules of [semantic version](https://semver.org/) we should move to a next major version. But since we are still on 0 that does not necessarily apply.
Since the package is already being used in production quite a lot I think a move to version 1.0.0 is warranted.

PR #64, opened 2017-11-24
Rename shed-install to shed-tools install and add --latest and --revisions flag. Add shed-tools update subcommand. 

rhpvorderman:
Solves #63 

Shed-install was diversified into an `install` and `update` command. The syntax now is changed.:
- To install:  `shed-install [flags]` -> `shed-tools install [flags]`
- To update: `shed-tools update`

This uses built in functionality:
- ~~`shed-install`~~  `shed-tools install` already defaulted to the latest installable revision when no revision is given for a tool. The `--latest` flag forces this behaviour.
- ~~`shed-install --update`~~ `shed-tools update` now uses get_tool_list_from_galaxy to get a tool list from the existing galaxy as a dictionary. Then it does the same as `shed-tools install --latest`.

To test all three possible methods of tool specifying for ~~shed-install~~ `shed-tools install`, `--revisions` was added to allow revisions installation from the command line. 

Some other changes had to be made to enable this.
- refactor ~~shed-install~~ shed-tools variables to be more readable
- move some common methods out of ~~shed-install~~ shed-tools to reduce the file length. This includes the logging
- write extra tests
- get_tool_list now does not write a file on instantiation anymore, but instead creates a class that can write a file or just simply give the dictionary.

Other nice changes that resulted from the refactoring:
- run-data-managers and ~~shed-install~~ shed-tools now use the same logging method. Logging can now also be extended to other tools.
- the `get_galaxy_connection` function was updated so it can optionally also use a file (i.e. a tool_list) to determine connection parameters.
- get_tool_list now uses bioblend api instead of custom connection.
- A check_url method checks the galaxy_url now for get_tool_list and run-data-managers as well broadening the scope of #49

EDIT: 2017-11-27, add changes to reflect further changes.


ephemeris/shed_install.py
```diff
@@ -229,29 +129,34 @@ def installed_tool_revisions(gi=None, omit=None):
...
-    tsc = tool_shed_client(gi)
+    tool_shed_client = ToolShedClient(gi)
+
+    # Create dictionary to look up all tools based on repository information
+
     installed_revisions_list = []
-    itl = tsc.get_repositories()
-    for it in itl:
-        if it['status'] == 'Installed':
+    installed_tool_list = tool_shed_client.get_repositories()
```
mvdbeek: Now that you're at it, can we change this to `installed_repository_list`, to avoid any confusion, and to leave some wigglespace should we ever implement a more tool-centric installation approach?

In fact I think you can s/tool/repository/g and s/tools/repositories/g in this file, this should be much clearer.

rhpvorderman:
@jmchilton It would be nice if it could be included in the following ephemeris release. #61 . After someone reviews it of course.

ephemeris/shed_install.py
rhpvorderman: On my way

rhpvorderman:
Done with all the refactoring.

rhpvorderman:
There is still some issue with log messages being displayed twice on shed-install

rhpvorderman:
Solved it

jmchilton:
Oh wow - this is fantastic!

ephemeris/shed_install.py
```diff
@@ -225,122 +139,128 @@ def installed_tool_revisions(gi=None, omit=None):
...
+                    in the repository not being included in the returned list.
     :rtype: dict
     :return: The returned dictionary contains the following keys, each
              containing a list of dictionaries:
-                - `tool_panel_shed_tools` with a list of tools available in the
+                - `tool_panel_shed_repositories` with a list of repositories available in the
                 tool panel that were installed on the target Galaxy instance
                 from the Tool Shed;
-                - `tool_panel_custom_tools` with a list of tools available in
+                - `tool_panel_custom_tools` with a list of repositories available in
```
bgruening: repositories should here be tool, right?

ephemeris/shed_install.py
```diff
@@ -225,122 +139,128 @@ def installed_tool_revisions(gi=None, omit=None):
...
-                 output of the `installed_tool_revisions` function, as
-                 `shed_tools` key).
+    .. seealso:: `installed_repository_revisions` (this function also returns the
+                 output of the `installed_repository_revisions` function, as
+                 `shed_repositories` key).
     """
     if not omit:
         omit = []
-    tp_tools = []  # Tools available in the tool panel and installe via a TS
+    tool_panel_repos = []  # The repositories from tools available in the tool panel and installable via a TS
```
bgruening: s/from/of/

maybe the 'the Tool Shed repositories of tools available ...'

ephemeris/shed_install.py
```diff
@@ -368,6 +288,13 @@ def _parser():
                         help="Galaxy tool panel section label where tool will be installed "
                              "(if the section does not exist, it will be created; "
                              "only applicable if the tools file is not provided).")
+    parser.add_argument("--revisions",
+                        default=None,
+                        dest="revisions",
+                        help="The revisions of the tool repository that will be installed. "
+                             "Revisions must be specified in on the command line in YAML format:"
```
bgruening: s/in//

But this is rather odd for the UX. Can we use `nargs=""` here instead?

ephemeris/shed_install.py
```diff
@@ -384,6 +311,16 @@ def _parser():
...
                              "Will be ignored on galaxy releases older than 16.07. "
                              "Can be overwritten on a per-tool basis in the tools file")
+    parser.add_argument("--latest",
+                        action="store_true",
+                        dest="force_latest_revision",
+                        help="Will override the revisions in the tools file and always install the latest revision.")
+    parser.add_argument("--update",
+                        action="store_true",
+                        dest="update_tools",
+                        help="This updates all tools in the Galaxy to the latest revision. "
```
bgruening: s/the//

bgruening: I'm wondering if we should make this a positional command.

Like `shed_tools update` and `shed_tools install ...`

tests/test.sh
```diff
@@ -14,25 +14,65 @@ workflow-install --help
...
+    CID=`docker run -d -e GALAXY_CONFIG_WATCH_TOOL_DATA_DIR=True -P bgruening/galaxy-stable`
+    # We get the webport (https://docs.docker.com/engine/reference/commandline/inspect/#list-all-port-bindings)
+    WEB_PORT=`docker inspect --format="{{(index (index .NetworkSettings.Ports \"$INTERNAL_EXPOSED_WEB_PORT/tcp\") 0).HostPort}}" $CID`
+    echo "Wait for galaxy to start"
+    galaxy-wait -g http://localhost:$WEB_PORT -v --timeout 120
+}
+
+function start_new_container {
+    echo "Start new container"
+    docker rm -f $CID
```
bgruening: Interesting will this automatically stop the container? Or should we stop the container before?

tests/test.sh
```diff
@@ -14,25 +14,65 @@ workflow-install --help
...
-echo "Test galaxy-wait function"
-galaxy-wait -g http://localhost:$WEB_PORT -v --timeout 120
+start_container
 docker ps
 
-echo "Check tool installation"
+echo "Check tool installation with yaml on the commandline"
+# CD Hit was chosen since it is old and seems to be unmaintained. Last update was 2015.
+# Anyone know a smaller tool that could fit its place?
+OLD_TOOL="{'owner':'jjohnson','name':'cdhit','revisions':['34a799d173f7'],'tool_panel_section_label':'CD_HIT'}"
```
bgruening: We should probably create a dummy tool in the TS. As soon as this container gets updated we will have failing tests that are hard to understand.

mvdbeek:
@rhpvorderman I added some minor tweaks here: https://github.com/rhpvorderman/ephemeris/pull/2

This works really well for me!

tests/test.sh
rhpvorderman: rm -f automatically stops the container. That is what the -f flag is for.

tests/test.sh
rhpvorderman: I agree. Better even to use the test toolshed for this.

ephemeris/shed_install.py
rhpvorderman: Yes. I did some find -> replace (in selection) stuff. It worked better than just doing it for the entire file. But I still missed some things. Thanks for pointing them out.

rhpvorderman:
Will now work on making update a subcommand.

rhpvorderman:
Done with everything except the dummy tool. I do not have time to work on this though. (I only have uploaded one tool in the toolshed and that was some time ago, so this will take quite some time for me to get right again) Can somebody else pick this up? Or we could file an issue and leave this for a later date.

I have altered the message at the top to reflect the further changes.

tests/test.sh
mvdbeek: No, this is fine, the test toolshed is not a good idea. We do similar stuff in galaxy, and even if the repository get's updated this will continue to work.

mvdbeek: actually we can just make sure that there is a new revision, we don't care which revision that is.

bgruening: @mvdbeek correct me, but if we test the `latest repository` as we do a few lines below, this test will fail as soon as the repo is updated.

mvdbeek: right, but we can test that there is an additional revision in the tool list

rhpvorderman: > right, but we can test that there is an additional revision in the tool list

Additional revisions are only installed if there are multiple versions. If the revision is of the same version, then it will override the previous revision and not show up with get-tool-list.

CD-HIT was chosen because it was last updated more than 2 years ago. The last update before that was 5 years ago. It seems to be pretty stable. If it changes, it is not too hard to figure out why the tests fail.

If we really want to do this right, we use a docker image to start up a custom toolshed, upload a custom tool to it, and then test it. But I am not sure this is worth the effort. Personally I think the test as it is now is sufficient.

bgruening:
This is a great Christmas present for the Galaxy community. Thanks a bunch @rhpvorderman!

xref: https://github.com/galaxyproject/ephemeris/issues/66

martenson:
Simply great, awesome contribution @rhpvorderman, thank you!

PR #70, opened 2018-01-03
safeguard revisions

bgruening:
`revisions` can be None ....


rhpvorderman:
```python
revisions = repo_info.get('revisions', [])
```
Should never return `None` right? In that case it should return an empty list `[]` which has length zero.

bgruening:
Not in this case:

```
>>> repo_info = {'foo': None}
>>> revisions = repo_info.get('foo', [])
>>> print revisions
None
```

PR #74, opened 2018-02-01
Makes publishing of imported workflows available.

pcm32:
Allows immediate publishing of loaded workflows.


mvdbeek:
Thanks a lot @pcm32 !

Issue #69, opened 2017-12-28
shed-tools update should include data managers


rhpvorderman:
This can be an optional thing.

By extension, this means get-tool-list should get functionality to also get the data-managers.

rhpvorderman:
Fixed by #78

PR #78, opened 2018-03-06
Also update data managers

rhpvorderman:
Hi. This solves #69 .
Have not run extensive testing yet, but will do that tomorrow.
It seems to run for now. Curious to see your feedback.

One of the side effects is that `get-tool-list` now also uses `get_galaxy_connection` method. a nice step towards unifying ephemeris.

EDIT:
Changes:
- `shed-tools update` now also updates data managers.
- With `get-tool-list` you can now also get all installed data managers, provided you have admin access to the galaxy.

### DIFF
--- a/ephemeris/__init__.py
+++ b/ephemeris/__init__.py
@@ -52,7 +52,7 @@ def load_yaml_file(filename):
     Load YAML from the `tool_list_file` and return a dict with the content.
     """
     with open(filename, 'r') as f:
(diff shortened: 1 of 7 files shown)


jmchilton:
Very nice - 👍 from me. Thanks a bunch @rhpvorderman !

rhpvorderman:
@bgruening @jmchilton Thanks for reviewing!

Issue #76, opened 2018-02-07
Project: Run tests on tools updates


jmchilton:
- [x] Update planemo to allow tool testing against external servers.
- [x] Develop tooling to automate this as part of ephemeris updates.
- [x] usegalaxy-eu integration @erasche 
  - [x] test usegalaxy-eu tools (https://github.com/usegalaxy-eu/usegalaxy-eu-tools).
  - [x] test usegalaxy-eu worflows

erasche:
Is there any possibility we could implement workflow testing while we're there? https://github.com/galaxyproject/galaxy/projects/8#card-7262571

EU really desperately wants this and we'll help however we can. (Need a crate of your favourite beverage/brand of hash browns? We can do this! Need a live popular server to test against? Done.) We would really love to be able to prove that all of the training workflows will run successfully without issue on our server.

rhpvorderman:
@jmchilton can this issue be closed? You successfully enabled tool testing from the command line.

erasche:
We haven't ticked the last box yet @rhpvorderman (or, second to last now that I've re-organised it.) We'll get there soon hopefully.

erasche:
I tried to run `planemo test --engine galaxy tools-iuc/tools/jq.xml` yesterday but it failed to load test data, it seemed to try and load that from the server path and not my local path.

mvdbeek:
Did you do it with an admin key ? AFAIK we didn't lift that requirement yet.

erasche:
Yup, did it under my admin account.

erasche:
```
urllib3.connectionpool DEBUG 2018-08-06 16:08:43,058: https://usegalaxy.eu:443 "GET /api/tools/jq/test_data_path?filename=list.json&key=<redacted> HTTP/1.1" 200 None
[Errno 2] No such file or directory: '/usr/local/galaxy/shed_tools/toolshed.g2.bx.psu.edu/repos/iuc/jq/5ff75eb1a893/jq/test-data/list.json'
Traceback (most recent call last):
  File "/home/hxr/arbeit/planemo/.venv/bin/planemo", line 11, in <module>
    load_entry_point('planemo', 'console_scripts', 'planemo')()
  File "/home/hxr/arbeit/planemo/.venv/lib/python3.5/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
[...]

PR #81, opened 2018-03-12
Option to test tools on update/install for Galaxy 18.05.

jmchilton:
Builds on #78 from @rhpvorderman since it modifies similar files.

While it does run the tests and produce a JSON output file that should be usable with ``planemo test_reports`` (http://planemo.readthedocs.io/en/latest/commands.html#test-reports-command) after the next Planemo release, marked as WIP because:

- [x] The tool tests create a user and a user API key to run tool tests, this should be configurable.
- [x] It doesn't have any in-ephemeris reporting yet - not display of test results. What do people want to see here? Probably the JSON is not enough?
- [x] Probably we should break up timing code also and have separate timing for install and testing?
- [x] Failed tests don't modify the ``shed-tools`` exit code.
- [x] ~Do we want to do all the installs and then all the tests like this or should we try to interleave them or should we test as we install or should be make this configurable?~ Lets just stick with doing them at the end and leave things be for now - we can revisit in the future.

Also, what of these questions can we delay to a second iteration?

Implements https://github.com/galaxyproject/ephemeris/issues/76.

### DIFF
--- a/ephemeris/shed_tools.py
+++ b/ephemeris/shed_tools.py
@@ -36,6 +36,8 @@ Galaxy's configuration directory and set Galaxy configuration option
 # bioblend, pyyaml
 
 import datetime as dt
+import json
+import re
 import sys
 import time
 from argparse import ArgumentParser
@@ -44,11 +46,12 @@ import yaml
 from bioblend.galaxy.client import ConnectionError
 from bioblend.galaxy.toolshed import ToolShedClient
 from bioblend.toolshed import ToolShedInstance
+from galaxy.tools.verify.interactor import GalaxyInteractorApi, verify_tool
 
 from . import get_galaxy_connection, load_yaml_file
 from .common_parser import get_common_args
 from .ephemeris_log import disable_external_library_logging, setup_global_logger
-from .get_tool_list_from_galaxy import GiToToolYaml
+from .get_tool_list_from_galaxy import GiToToolYaml, tools_for_repository
 
 # If no toolshed is specified for a tool/tool-suite, the Main Tool Shed is taken
 MTS = 'https://toolshed.g2.bx.psu.edu/'  # Main Tool Shed
@@ -65,6 +68,8 @@ MTS = 'https://toolshed.g2.bx.psu.edu/'  # Main Tool Shed
 INSTALL_TOOL_DEPENDENCIES = False
 INSTALL_REPOSITORY_DEPENDENCIES = True
 INSTALL_RESOLVER_DEPENDENCIES = True
+EXIT_CODE_INSTALL_ERRORS = 1
+EXIT_CODE_TOOL_TEST_ERRORS = 2
 
 
 def _ensure_log_configured(name):
@@ -275,6 +280,7 @@ def _parser():
     # update_tools in the name space and shed-tool update will not return all the install
     # variables.
     parser.set_defaults(
+        action="install",
         tool_list_file=None,
         tool_yaml=None,
         owner=None,
@@ -285,7 +291,12 @@ def _parser():
         tool_shed_url=None,
         skip_tool_dependencies=False,
         install_resolver_dependencies=False,
-        force_latest_revision=False
+        force_latest_revision=False,
+        test=False,
+        test_user_api_key=None,
+        test_user="ephemeris@galaxyproject.org",
+        test_json="tool_test_output.json",
+        test_existing=False,
     )
     install_command_parser = subparsers.add_parser(
         "install",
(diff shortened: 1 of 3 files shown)


rhpvorderman:
@jmchilton can you rebase on master? The commits from #78 are now in.

ephemeris/shed_tools.py
```diff
@@ -363,6 +369,27 @@ def _parser():
         help="This updates all tools in Galaxy to the latest revision."
              "Use shed-tools update --help for more information",
         parents=[common_arguments])
+
+    for command_parser in [update_command_parser, install_command_parser]:
+        command_parser.add_argument(
+            "--test",
+            action="store_true",
+            dest="test",
+            help="Run tool tests on install tools, requires Galaxy 18.05 or newer."
```
rhpvorderman: This requirement should be hardcoded somewhere
`gi.config.get_config()["version_major"]` returns the version.

ephemeris/shed_tools.py
```diff
@@ -363,6 +369,27 @@ def _parser():
...
+        )
+        command_parser.add_argument(
+            "--test_json",
+            dest="test_json",
+            help="If --test is specified, record tool test output to specified file. "
+                 "This file can be turned into nice reports with ``planemo test_reports <output.json>``."
```
rhpvorderman: Personal taste: I would leave out the word 'nice' here.

ephemeris/shed_tools.py
```diff
@@ -363,6 +369,27 @@ def _parser():
...
+        )
+        command_parser.add_argument(
+            "--test_existing",
+            action="store_true",
+            help="If --test is specified, also run tool tests on repositories already installed "
+                 "(i.e. skipped repositories)."
+        )
+
```
rhpvorderman: Cool!

ephemeris/shed_tools.py
```diff
@@ -657,7 +694,53 @@ def install_repositories(self):
...
+            for installed_repository in target_repos:
+                repo_tools = tools_for_repository(self.gi, installed_repository)
+                installed_tools.extend(repo_tools)
+
+            test_results = []
+
+            for tool in installed_tools:
+                galaxy_interactor_kwds = {
+                    "galaxy_url": re.sub('/api', '', self.gi.url),
+                    "master_api_key": self.gi.key,
```
rhpvorderman: Will self.gi.key always return the key? Also if initiated with email and password. I tested it and it seems to be the case. But is this certain?

ephemeris/shed_tools.py
```diff
@@ -657,7 +694,53 @@ def install_repositories(self):
                 t.get('changeset_revision', "")
             ) for t in self.errored_repositories])
         )
-        log.info("All repositories have been processed.")
+        log.info("All repositories have been installed.")
+        if self.test:
+            installed_tools = []
```
rhpvorderman: Maybe this should get its own method? `test_repositories"

ephemeris/shed_tools.py
```diff
@@ -657,7 +694,53 @@ def install_repositories(self):
...
+            if self.test_existing:
+                target_repos += self.skipped_repositories
+            for installed_repository in target_repos:
+                repo_tools = tools_for_repository(self.gi, installed_repository)
+                installed_tools.extend(repo_tools)
+
+            test_results = []
+
+            for tool in installed_tools:
+                galaxy_interactor_kwds = {
```
rhpvorderman: This whole code block can be in a `test_tool` method. This makes it a lot cleaner:
```python
for tool in installed_toos:
  self.test_tool(tool)

rhpvorderman:
I am all for integrating the testing within the installation method. But this causes indeed the exit code problem. Doing two things at the same time creates ambiguity.
However, you could argue that if a test fails, this means the tool should not be installed, which means installation has effectively failed.
Also I am in favor of adding a `shed-tools test` subcommand, which can be regularly by cron, to check if all the installed tools still functions.

ephemeris/shed_tools.py
jmchilton: I don't know about this - Nate had mentioned wanting to get this on usegalaxy.org and we were intending to maybe place the Galaxy-side APIs in the usegalaxy branch to allow this testing against our variant of 18.01. I guess we could just hack up ephemeris though if we do get around to setting this up. I'll try to find a spot to make this change.

ephemeris/shed_tools.py
jmchilton: https://github.com/galaxyproject/bioblend/blob/master/bioblend/galaxyclient.py#L202

Yes - as long as ``login_required=True`` with ``get_galaxy_connection`` - which it is for this command. It'd be nice if galaxy-lib had a variant of the client code that consume and used a bioblend GalaxyInstance but it doesn't yet so I think this is okay for now? I will admit this is a hack though - let me know if you need me to push the hack down a layer into galaxy-lib.

ephemeris/shed_tools.py
jmchilton: Agreed - I'll try to make this change.

ephemeris/shed_tools.py
jmchilton: Agreed - I'll try to make this change.

jmchilton:
> However, you could argue that if a test fails, this means the tool should not be installed, which means installation has effectively failed.

Yes - I think I agree. I'll see if I can make the exit code reflect the testing status.

> Also I am in favor of adding a shed-tools test subcommand, which can be regularly by cron, to check if all the installed tools still functions.

Absolutely - this is an excellent idea - I'll try to reorganize this code to do that.

PR #89, opened 2018-05-07
Upload 2.0 support for data library creation.

jmchilton:
Will convert older style library YAML format from ephemeris to modern version, will attempt to implement a subset of the modern format with legacy APIs as needed.

Some of the defaults that only made sense for training data have been hidden behind a new ``--training`` flag.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,6 +13,5 @@ logging-level=INFO
 [flake8]
 max-line-length = 150
 max-complexity = 30
-ignore = E501
 import-order-style = smarkets
 application-import-names = ephemeris
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,5 +13,6 @@ logging-level=INFO
 [flake8]
 max-line-length = 150
 max-complexity = 30
+ignore = E501
 import-order-style = smarkets
 application-import-names = ephemeris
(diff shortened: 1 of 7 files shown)


bgruening:
If we really want to support the old format as well we should add a test with the old file to https://github.com/galaxyproject/ephemeris/blob/master/tests/test_workflow_and_data.sh#L20

bgruening:
* legacy mode of new and old yaml file puts file into a lib with names like `https://raw.githubusercontent.com/yhoogstrate/segmentation-fold/55d0bb28b01e613844ca35cf21fa41379fd72770/scripts/energy-estimation-utility/tests/test-data/workflow-test_cd-box_kturns.xml` --> can we use here just the filename?

jmchilton:
>  If we really want to support the old format as well we should add a test with the old file to https://github.com/galaxyproject/ephemeris/blob/master/tests/test_workflow_and_data.sh#L20

Yeah - I added a file to do that but forgot to update the test. I'll do that.

setup.cfg
```diff
@@ -13,5 +13,6 @@ logging-level=INFO
 [flake8]
 max-line-length = 150
 max-complexity = 30
+ignore = E501
```
rhpvorderman: Long URLs are ignored by standard. So I cannot see a reason why line length should be ignored.

rhpvorderman:
Pull request looks good. Travis only fails on a few linting errors.

jmchilton:
> legacy mode of new and old yaml file puts file into a lib with names like https://raw.githubusercontent.com/yhoogstrate/segmentation-fold/55d0bb28b01e613844ca35cf21fa41379fd72770/scripts/energy-estimation-utility/tests/test-data/workflow-test_cd-box_kturns.xml --> can we use here just the filename?

Added an example of how to do that and explained why I'd prefer not to make it automatic in this commit's comment:

https://github.com/galaxyproject/ephemeris/pull/89/commits/dea1ff2f9ff94ef1d3df16582d8245b365bf31dd

bgruening:
Thanks @jmchilton!

PR #90, opened 2018-05-08
Fix a few lint issues.

jmchilton:


setup.cfg
```diff
@@ -13,5 +13,6 @@ logging-level=INFO
 [flake8]
 max-line-length = 150
 max-complexity = 30
+ignore = E501
```
rhpvorderman: Why ignore line too long? Or does this interfere with max-line-length?

bgruening: We ignore this on most Galaxy project. I'm just wondering why we get more lint errors it we ignore E501. Before I added this it was only E501 error, now its much more.

rhpvorderman: Why ignore it, and not fix the line length instead?

rhpvorderman: In my opinion it is better to strictly enforce style (including line length) in projects that have multiple users collaborating, because this improves readability. But in the end this is not my decision to make. My concern is that not strictly enforcing style will lead to maintainability issues down the road.

bgruening: Everyone agree on this. But a lot of people do not follow this specific rule as we all have big monitors :) 
https://github.com/search?q=ignore+%3D+E501&type=Code

bgruening: See also here: https://galaxyproject.org/develop/best-practices/

jmchilton: It looks like Pulsar and Planemo enforce the max line length of 150 also, I think agree that should be long enough. I'd prefer just to fix the long lines. I thought I put this in there because it was enforcing 80 for some reason which is in my opinion unworkable when I've tried it in the past. Mind if I close this PR @bgruening, sorry I was confused.

bgruening: please go on ...

drosofff: @jmchilton Please feel free to remove 

'''
                                   "Christophe Antoniewski <drosofff@gmail.com>\n"
                                   "https://github.com/ARTbio/ansible-artimed/tree/master/extra-files/generate_tool_list_from_ga_workflow_files.py")
```
from generate_tool_list_from_ga_workflow_files.py. It is non sense as well as deprecated.

PR #93, opened 2018-05-10
WIP: enhance shed_tools to install from native data_manager files

bgruening:


ephemeris/shed_tools.py
```diff
@@ -579,6 +583,20 @@ def get_install_repository_manager(options):
             'install_tool_dependencies', INSTALL_TOOL_DEPENDENCIES)
     elif options.tool_yaml:
         repositories = [yaml.safe_load(options.tool_yaml)]
+    elif options.data_manager_yaml:
+        repositories = list()
+        dms = yaml.load(open(options.data_manager_yaml))
+        for dm in dms['data_managers']:
+            tokens = dm['id'].split('/')
```
rhpvorderman: In a data managers file you can also just give the tool ID so then this does not work, and a lot of errors will follow...
There should be some error handling here. Also tests should be added to include a parsable file and a non-parsable (but still valid) file.

rhpvorderman: The data manager job is saved in galaxy's history. So that is not necessary IMO.Also, long tool ids are quite inconvenient. Are there any people who have multiple fasta indexes of the same genome on their galaxy just because of the version of their indexer?

bgruening: Oh its not about reproducibility in one Galaxy instance, but across instances... https://github.com/bgruening/idc

> Are there any people who have multiple fasta indexes of the same genome on their galaxy just because of the version of their indexer?

Not sure, but they are different data managers where the version is crucial. And keep in mind that data managers can be way more than genome indices.

rhpvorderman: These are valid reasons to support long tool IDs but not necessarily to enforce them.
Would it be possible to put in an if statement here that recognizes a long tool ID from a short tool ID and else throws a proper exception? For the long tool ID the following code looks good. For a short tool ID some code could be added that fetches the repo owner from toolshed, preferring `iuc` and `devteam` if there are multiple owners and selecting the first one in alphabetic order otherwise whilst throwing a warning that there were multiple owners.

Issue #91, opened 2018-05-09
Project: refactor shed-tools


rhpvorderman:
* it's 1000 lines of code
* It contains multiple functions that are over 50 lines long.
* It can contains a class (that should be able to simplify a few things) but in this class multiple methods are over 50 lines long.
* A global log variable is defined
* It overwrites the ConnectionError that is in standard library.
* It has by far the highest complexity of all files in ephemeris (18)

Shed-tools is not pleasant to work at. Navigating through it is a jungle. It should be totally refactored.
I have a few suggestions how to approach this:
* Take `get_tool_list_from_galaxy.py` as an example. This was done for `run-data-managers` as well, and as a result complexity was dropped, while at the same time increasing functionality.
Essentially this means shed_tools should be written as a class that:
    - Has a method to output currently installed tools
    - Has a method to check whether a tool is currently installed
    - Has a method to install a tool
    - Has a method to test a tool (Thanks a lot @jmchilton !)
    - Has some sort of install_tools method for multiple tools including logging.
* The parsing of the tool list yaml may be moved to a separate file
* Some stuff from get_tool_list from galaxy might be reused in shed-tools. (Like determining current tool list)
* Some redundant command line options should be removed. Installing by passing a yaml string? OK. Installing by tool_list file? OK. Installing by specifying every frigging tool attribute on the commandline? We can drop that IMO.

This is something I have been thinking about doing it for a while. Just putting it out there before we increase the complexity of shed-tools even further. I think refactoring will also improve the ability to test using pytest #80.

rhpvorderman:
Done in #104

Issue #80, opened 2018-03-12
Change test framework


rhpvorderman:
Currently we use a bash script that calls the tools.
This is not bad. But the testing is quite crude and can never cover most use cases.
In order to solve #75 I think it is better to use pytest or something comparable. Just like [ncbi-genome-download](https://github.com/kblin/ncbi-genome-download/tree/master/tests).
That will probably also allow for better coverage integration  (#71 ).

rhpvorderman:
Test framework for pytest has been set up in #105. 
Now we need to port all the tests. But that is a different issue #75 .

Issue #101, opened 2018-07-02
shed-tools install not installing all dependencies


dwightkuo:
```shed-tools install -g http://127.0.0.1:8080/ -a 9da1b0b76e038a2a70e0649adb5205fa -t galaxy.yml --install_resolver_dependencies --latest --log_file logfile.txt```

>tools:
>\- name: tophat_fusion_post
>  owner: devteam
>  revisions:
>  \- f83394a2c2da
>  tool_panel_section_id: ngs_mapping
>  tool_panel_section_label: 'NGS: Mapping'
>  tool_shed_url: https://toolshed.g2.bx.psu.edu/

When I click on "Manage Tools", it shows "Installed, missing tool dependencies". When checking these, it seems many 'Installed repository dependencies' and 'Missing tool dependencies' have not been installed
![snapshot3](https://user-images.githubusercontent.com/22648143/42185533-159d904c-7dfe-11e8-9fe3-200317cee7fc.png)
. Is this an issue or am I misunderstanding how to use shed-tools?

rhpvorderman:
This is an issue. 
Galaxy has two sorts of dependencies: Those that are installed using `conda` and those that are installed via the galaxy package management system. The latter are being phased out. The default behavior of `shed-tools` is therefore not to install tool-dependencies. Unfortunately this cannot be changed from the command line.

I will create a few commits in #104 to alleviate the issue.

rhpvorderman:
This should be fixed with #104. 

Can you run ephemeris from the develop version?
- `git clone https://github.com/galaxyproject/ephemeris.git`
- Go to ephemeris directory
- `git checkout master`
- `git pull`
- `python setup.py install`

You are now running ephemeris development version. `shed-tools` should be available.
Can you install tool dependencies now?

dwightkuo:
Yes, the development version fixes the issue!

rhpvorderman:
I am happy to hear that. The next release of ephemeris will have these fixes incorporated. But until that time the development version should work fine.

PR #104, opened 2018-07-18
Done: Refactor shedtools

rhpvorderman:
~~After two attempts refactoring the existing file, I am going to refactor shed tools from scratch. I will use the issue #91 as a guideline.~~

~~Please do not add any functionality to shed tools before the refactoring complete. I will be pushing regularly so you can check on my progress. I will also be available on gitter.~~

Done.
The following stuff has changed:
+ shed_tools arguments and methods now in a separate file (for readability reasons)
+ InstallToolManager rewritten to allow for a much cleaner interface. No storing of variables in self that are used by only one method. InstallToolManager can now be instantiated with only a GalaxyInstance. Methods also have a sane number of arguments. So, when people where referencing shed_tools.py in other scripts, these scripts are now broken. On the upside, referencing shed_tools in other scripts is now much easier
+ Global log was removed
+ Complexity dropped by 3 units. (Whatever that is supposed to mean)
+ Useless stuff was chucked out. Some defaults were passed through functions 5 times.
+ get-tool-list can now get all the revisions of each tool, not just the revisions that are in the tool panel. Admin rights are needed for this. Shed-tools uses this functionality to get the installed tools.
+ Whether tools are already installed is now determined by getting a list of the installed tools and checking that. The previous method did this via checking every tool with an API call to galaxy. This was of course slow. Now it should be faster (even though the checking method is gruesomely inefficient, but it should be faster than doing ssl handshakes all the time).
+ `shed-tools update` now also accepts tool lists. It will only update tools that are already installed on the galaxy and issue a warning for other tools.
+ The `--skip_install_tool_dependencies` flag is no longer functional as this is the default. Same goes for the `--install_resolver_dependencies` flag. These flags are still around for backwards compatibility, but do not do anything. `--install_tool_dependencies`, `--skip_install_resolver_dependencies` and `--skip_install_repository_dependencies` where added.
+ Probably some stuff I forgot.

The following stuff has not changed:
+ The stuff that shows up on your terminal when installing. Kept that exactly the same. 

Stuff that needs further disccussion:
~~+ The CLI interface has some mentions of skipping tool dependencies and installing resolver dependencies. Skipping tool dependencies and installing resolver and repository dependencies are now the default, so these options do not make sense. They are ignored in `def main` but kept for backwards compatibility. (I have broken every one's scripts already once... )~~

I hope it is much easier to work now on `shed_tools`. Should also be a lot easier to write a testing framework now using pytest #80 .

Hope this somewhat redeems myself for not being at the galaxy hackathon.


tests/data_manager_list.yaml
```diff
@@ -7,4 +7,6 @@ tools:
   owner: devteam
 - name: data_manager_bwa_mem_index_builder
   owner: devteam
+  revisions:
+    - e519769cdb9c
```
mvdbeek: Can you try 46066df8813d as well, which is the latest version? I just tested this revision and it works, my hypothesis is that not specifying a revision doesn't install the latest revision after these changes.

rhpvorderman: Yeah sure, will try that.

mvdbeek: Do you have the log / commit somewhere ?

rhpvorderman:
Added extra CLI functionality to solve #101

mvdbeek:
I think I have implemented all those changes in https://github.com/rhpvorderman/ephemeris/pull/3

rhpvorderman:
Thanks for all the changes @mvdbeek. Especially for fixing the error of get-tool-list listing deleted tools. I ran into this error, but did not really understand why it was happening, and then forget about it. Thanks for all the effort on this pull request!

> This is nice, I like it in principle. I'm ambivalent on the name change of install_repositories to install_tools, because we only install repositories, and I'm afraid that distinction is lost with these changes. Which is sort-of OK, because nothing here actually deals with tools, and we were not very consistent from the start. 

Yes this is quite ambivalent trough ephemeris. `get-tool-list` also gives you a list of repositories. So making this consistent across ephemeris would mean we have to use `get-repository-list`. Internally it can be made consistent though, to avoid confusion. 

Since we are going to break planemo anyway, I better do that now to avoid breaking it twice. Commits coming up.

rhpvorderman:
Renamed all the tools that are actually repos to repos.

I kept the `test_tools` method, since it actually tests tools. It tests all the tools in a certain list of repos. So kept that the same. I hope it is not too confusing.

Also kept the user interface the same. I can rename everything in the documentation to repositories properly, and also change the information in the parser. But I do not think that is something we should do in this pull request.

.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mvdbeek: We'll want to put this file into .gitignore, right ?

rhpvorderman: Whoops. Did put this in global .gitignore. But apparently got committed anyway.

rhpvorderman: fixed now by recommitting changes

PR #106, opened 2018-08-15
fix coverage reporting on codacy

rhpvorderman:
Apparently the setup was not working. It appears that codacy needs to know the exact path of the source files for coverage to work. By contrast, coverage with pytest in a tox environment only works when the package site is used to determine coverage. 

Luckily `sed` can rescue us in this case.
I've tested this manually, and coverage reports are accepted now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -41,9 +41,16 @@ commands = pytest {[tox]test_dir}
 
 [testenv:py35-pytest]
 passenv = TRAVIS,CODACY_PROJECT_TOKEN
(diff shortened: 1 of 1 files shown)


rhpvorderman:
Coverage results can be checked here:

https://app.codacy.com/project/galaxyproject/ephemeris/dashboard

bgruening:
Pretty cool! Thanks a ton for all the work which you have put into this project this is awesome!

PR #108, opened 2018-08-16
Log stderr of a data manager job failing.

### LINKED ISSUE #107: Difficult to debug run_data_managers
When running the data manager from ephemeris, there is no way to see what went wrong unless you tell galaxy to turn off cleanup and go digging in the job working directory. This is very awkward to debug if it is being run during a docker build, say, after the tools install. Here is what one gets in the case of any error:

```
galaxy.jobs.output_checker DEBUG 2018-08-15 19:49:06,367 Tool produced standard error failing job - [Traceback (most recent call last):
  File "/shed_tools/toolshed.g2.bx.psu.edu/repos/trinity_ctat/ctat_genome_resource_libs_data_manager/ea7bc21cbb7a/ctat_genome_resource_libs_data_manager/data_manager/add_ctat_resource_lib.py", line 879, in <module> # this is the line that says main(), not helpful in this case..
]
galaxy.jobs DEBUG 2018-08-15 19:49:06,526 (2) setting dataset 2 state to ERROR
galaxy.jobs INFO 2018-08-15 19:49:06,636 Collecting metrics for Job 2
galaxy.jobs DEBUG 2018-08-15 19:49:06,664 job 2 ended (finish() executed in (162.449 ms))
galaxy.tools.error_reports DEBUG 2018-08-15 19:49:06,674 Bug report plugin <galaxy.tools.error_reports.plugins.sentry.SentryPlugin object at 0x7f1a1e525190> generated response None

==> /home/galaxy/logs/slurmd.log <==
[2018-08-15T19:49:06.299] [3] sending REQUEST_COMPLETE_BATCH_SCRIPT, error:0
[2018-08-15T19:49:06.306] [3] done with job
Job 2 finished with state error.
Not all jobs successful! aborting...
Traceback (most recent call last):
  File "/usr/local/bin/run-data-managers", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/dist-packages/ephemeris/run_data_managers.py", line 294, in main
    data_managers.run(log, args.ignore_errors, args.overwrite)
  File "/usr/local/lib/python2.7/dist-packages/ephemeris/run_data_managers.py", line 255, in run
    run_jobs(self.fetch_jobs, self.skipped_fetch_jobs)
  File "/usr/local/lib/python2.7/dist-packages/ephemeris/run_data_managers.py", line 248, in run_jobs
    raise Exception('Not all jobs successful! aborting...')
Exception: Not all jobs successful! aborting...
```
Is there a way to display the contents of the galaxy_#.e file, perhaps?

rhpvorderman:
Added the code to log a stderr when a data manager job is failing.

I also ported `run-data-managers` to the new pytest framework. So this extra logging stuff is also tested.

This should fix #107 .


rhpvorderman:
I have changed the image as the `bgruening/galaxy-stable:dev` was giving problems. `bgruening/galaxy-stable:latest` seems to be a better choice.

tests/test_run_data_managers.py
```diff
@@ -0,0 +1,90 @@
...
+            dict(name="data_manager_fetch_genome_dbkeys_all_fasta",
+                 owner="devteam"),
+            dict(name="data_manager_sam_fasta_index_builder",
+                 owner="devteam"),
+            dict(name="data_manager_bwa_mem_index_builder",
+                 owner="devteam")
+        ]
+        irm = InstallRepositoryManager(container.gi)
+        irm.install_repositories(data_managers)
+        print("Restart container")
```
mvdbeek: Is that necessary ? It should work without on the 18.05 image, otherwise that'd be a galaxy bug.

tests/test_run_data_managers.py
```diff
@@ -0,0 +1,90 @@
...
+                    params=[
+                        {'dbkey_source|dbkey_source_selector': 'new'},
+                        {'dbkey_source|dbkey': 'INVALID'},
+                        {'dbkey_source|dbkey_name': 'INVALID_KEY'},
+                        {'sequence_name': 'INVALID'},
+                        {'sequence_id': 'INVALID'},
+                        {'reference_source|reference_source_selector': 'ncbi'},
+                        {'reference_source|requested_identifier': 'INVALID0123'}
+                    ],
+                    data_table_reload=[
```
mvdbeek: again, that shouldn't be necessary, and we'll remove the explicit table reloads in the future, as they don't even work in most setups.

tests/test_run_data_managers.py
rhpvorderman: I copied this from the old test script (where the container needed to be restarted after new data managers are installed). But I will remove this.

tests/test_run_data_managers.py
rhpvorderman: `data_table_reload` is used to determine which table needs to be populated. It is necessary for the checking functions. So it is still used, but not for the functionality it was originally meant for.
This might be more elegantly fixed since this is an attribute of the datamanager anyway, so the users should not have to supply this. But let's do that in a later pull request.

tests/test_run_data_managers.py
rhpvorderman:
```
    def make_post_request(self, url, payload, params=None, files_attached=False):
        """
            Make a POST request using the provided ``url`` and ``payload``.
            The ``payload`` must be a dict that contains the request values.
            The payload dict may contain file handles (in which case the files_attached
            flag must be set to true).
    
            If the ``params`` are not provided, use ``default_params`` class field.
            If params are provided and the provided dict does not have ``key`` key,
            the default ``self.key`` value will be included in what's passed to
            the server via the request.
    
            :return: The decoded response.
            """
        if params is not None and params.get('key', False) is False:
            params['key'] = self.key
        else:
            params = self.default_params
    
        # Compute data, headers, params arguments for request.post,
        # leveraging the requests-toolbelt library if any files have
        # been attached.
        if files_attached:
            payload.update(params)
            payload = MultipartEncoder(fields=payload)
            headers = self.json_headers.copy()
            headers['Content-Type'] = payload.content_type
            post_params = {}
        else:
            payload = json.dumps(payload)
            headers = self.json_headers
            post_params = params
    
        r = requests.post(url, data=payload, headers=headers,
                          verify=self.verify, params=post_params,
                          timeout=self.timeout)
        if r.status_code == 200:
            try:
                return r.json()
            except Exception as e:
                raise ConnectionError("Request was successful, but cannot decode the response content: %s" %
                                      e, body=r.content, status_code=r.status_code)
        # @see self.body for HTTP response body
        raise ConnectionError("Unexpected HTTP status code: %s" % r.status_code,
>                             body=r.text, status_code=r.status_code)
E       bioblend.ConnectionError: Unexpected HTTP status code: 400: {"err_data": {"all_fasta_source": "Parameter all_fasta_source requires a value, but has no legal values defined."}, "err_msg": "Parameter all_fasta_source requires a value, but has no legal values defined.", "err_code": 0}
```

rhpvorderman: Leads to an error unfortunately.
Let's solve this in another PR

mvdbeek: Yeah, I looked at the old test and I remember why we did this.
Can you add the comment `#We restart galaxy because otherwise the data manager tables won't be watched`

rhpvorderman:
So much clutter. And travis is giving issues with the environment. Furthermore codacy integration has been updated. I will close this and reopen in a new request.

PR #112, opened 2018-09-13
Prepare next release.

jmchilton:


rhpvorderman:
Sometimes jobs randomly fail for some reason. This is mainly due to the data manager part. Apparently sometimes the installation is not successful and the data managers need to be run again. I have never had this happen outside of the testing though. So I cannot really debug it. 
I restarted the failing job on travis.

Good work @jmchilton !

HISTORY.rst
```diff
@@ -6,10 +6,25 @@ History
...
----------------------
-
-    
+0.10.0.dev0
+---------------------
+
+* Refactor shed tool functionality to be more modular, readable, and concise
+  (thanks to `@rhpvorderman`_).
+  `Pull Request 104`_
+* Add ``pytest``, enable coverage testing (thanks to `@rhpvorderman`_).
```
rhpvorderman: Should all the testing be in the release note? I mean, it matters to us but what does the end user notice? (well less regressions probably, but we know that because we are developers. Our end users are not necessarily developers).

HISTORY.rst
```diff
@@ -6,10 +6,25 @@ History
...
 
 ---------------------
-0.9.1.dev0
----------------------
-
-    
+0.10.0.dev0
+---------------------
+
+* Refactor shed tool functionality to be more modular, readable, and concise
```
rhpvorderman: Maybe rephrase this to the things that matter to the end user:

Removed deprecated options from shed-tools CLI. shed-tools update now also accepts tool list, so tools in galaxy can be selectively updated. 
Improved algorithm leads to much faster skipping of already installed tool, which makes the installation of tools much faster on an already populated galaxy.

HISTORY.rst
martenson: @rhpvorderman I don't see reason why not to include it. Release notes have various people reading them and one of the reasons for notes is also to give credit (in this case to you ;)

HISTORY.rst
martenson: @rhpvorderman I am sure @jmchilton does not mind if you re-phrase this to your liking with a direct commit. Please do so.

rhpvorderman: @martenson Thanks for the suggestion. I changed it.

PR #118, opened 2019-01-03
Test tools in parallel, with regular user permissions, without a shared filesystem

mvdbeek:
~~We'll need a galaxy-lib update with https://github.com/galaxyproject/galaxy/pull/7167 for re-using the history.~~

We've got a new galaxy-lib update, so let's get this in. It allows testing tools without an admin API key and without a shared filesystem.

### DIFF
--- a/src/ephemeris/shed_tools.py
+++ b/src/ephemeris/shed_tools.py
@@ -234,8 +234,10 @@ class InstallRepositoryManager(object):
         all_test_results = []
 
         for tool in installed_tools:
+            log.info("Testing tool '%s'", tool['id'])
             results = self._test_tool(tool, test_user, test_user_api_key)
             all_test_results.extend(results.tool_test_results)
+            log.info("%s test passed, %d tests failed for tool '%s'" % (len(results.tests_passed), len(results.tests_exceptions), tool['id']))
             tests_passed.extend(results.tests_passed)
             test_exceptions.extend(results.test_exceptions)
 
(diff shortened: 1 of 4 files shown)


src/ephemeris/shed_tools.py
```diff
@@ -232,75 +235,103 @@ def test_tools(self,
...
+        return galaxy_interactor
+
+    @staticmethod
+    def _test_tool(executor,
+                   tool,
+                   galaxy_interactor,
+                   test_history=None,
+                   log=None,
+                   tool_test_results=None,
+                   tests_passed=None,
```
rhpvorderman: I don't see `tests_passed` and `test_exceptions` defined. Yet the code appends to them whether they are lists. Shouldn't these be empty lists then by default?

mvdbeek: This is a private method that is the target function for the ThreadPoolExecutor, to be called from `test_tools`. I can make these non-default, which makes some sense. `tests_passed ` and `test_exceptions `  are defined in `test_tools`.

rhpvorderman: :+1: Then better make them non-default indeed. The error will be less weird if anything goes wrong.

mvdbeek:
Alright, test are done and passing at https://travis-ci.org/galaxyproject/ephemeris/jobs/515746753

rhpvorderman:
Looks great!

rhpvorderman:
Thanks a lot @mvdbeek for implementing this!

mvdbeek:
Thanks for the review and merge!

PR #119, opened 2019-01-07
Fix  crash when tool_with_panel is empty

jvanbraekel:
Before : Crash when 'tool_with_panel' is empty and ToolShedClient.get_repositories() is not. 
After : Define tool_panel_section_id/label outside of the for loop to remove the conditional definition dependency with 'tool_with_panel'


rhpvorderman:
All tests succeed! Could you fix the whitespace issue? Then I am happy to merge.

jvanbraekel:
Yes, no clue why this change appends at the first place...

PR #122, opened 2019-02-11
Update Docs for User Name (Should be Email)

rdvelazquez:
This tripped me up when trying to use the `shed-tools install` command. I entered `-u rdvelazquez` which is my public user name on the instance but the command failed. I had to enter `-u <my email address>` for the command to work.


afgane:
Thanks! Yes, better to be more specific.

PR #123, opened 2019-03-15
Handle timeout gracefully for UWSGI connection

pcm32:
I don't really know what I'm doing here, but following @mvdbeek suggestion :-).

### DIFF
--- a/src/ephemeris/get_tool_list_from_galaxy.py
+++ b/src/ephemeris/get_tool_list_from_galaxy.py
@@ -103,7 +103,9 @@ class GiToToolYaml:
         if self.get_data_managers:
             for tool in self.installed_tool_list:
                 if tool.get("model_class") == 'DataManagerTool':
-                    repositories.append(get_repo_from_tool(tool))
+                    repo = get_repo_from_tool(tool)
+                    if repo:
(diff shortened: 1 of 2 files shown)


src/ephemeris/get_tool_list_from_galaxy.py
```diff
@@ -103,7 +103,9 @@ def record_repo(tool_elem):
         if self.get_data_managers:
             for tool in self.installed_tool_list:
                 if tool.get("model_class") == 'DataManagerTool':
-                    repositories.append(get_repo_from_tool(tool))
+                    repo = get_repo_from_tool(tool)
+                    if repo:
```
rhpvorderman: This is a good idea if a tool does not have a repo. Could you add a comment here with an example of why this is needed? For the people who read the code in the future and do not know all galaxy specifics by heart.

mvdbeek:
The download from NCBI doesn't seem to work currently, that may be an intermittent issue since it worked on Friday.

rhpvorderman:
Approved. The tests are running again if the issue was intermittent the PR should be ready to merge in a few minutes.
Thanks a lot @pcm32 and @mvdbeek for fixing this!

PR #128, opened 2019-04-25
Add pysam and continue if test fetching errors

mvdbeek:


rhpvorderman:
Pysam? Do we want ephemeris to have anything to do with bioinformatics content? Isn't that out of scope for this project?

mvdbeek:
Ephemeris does tool testing, and this is needed to test that outputs match.

rhpvorderman:
But do we then eventually not also need Biopython and cyvcf2 etc? I

mvdbeek:
No, this is only needed so that we can verify the bam and cram files match. Most other formats are text or compressed text based, no special tools needed there. I can move it to `extras_requires` though, then you'll not get it by default. If we don't have pysam and we run tool tests we can log a warning.

rhpvorderman:
>  No, this is only needed so that we can verify the bam and cram files match. Most other formats are text or compressed text based, no special tools needed there. I can move it to extras_requires though, then you'll not get it by default. If we don't have pysam and we run tool tests we can log a warning.

Thank you for explaining. I was worried that a lot of dependencies would be added to ephemeris, which could give problems down the road. But BAM files are indeed  one of the few things that cannot be tackled by the standardlib.  The extras_requires seems a good solution. On the other hand I think you are in a much better position to judge. If this is needed for a lot of tool tests then it is probably good to have this by default.

I just did a check on pysam, and it only depends on cython, which depends on nothing. So adding it as a default dependency does not create dependency hell for tools which depend on ephemeris. The only one that I know of is planemo, and that probably needs the tool testing functionality.

martenson:
@mvdbeek ready to take this out of draft?

martenson:
This is mostly superseded by https://github.com/galaxyproject/ephemeris/pull/137

The pysam requirement can and probably should be managed outside of Ephemeris. Thanks for the insights @rhpvorderman

mvdbeek:
Not sure why this was closed, we should still do it.

rhpvorderman:
I disabled the codacy check. It is slow and not really necessary since we can lint on the actions CI.

mvdbeek:
Thank you @rhpvorderman!

PR #130, opened 2019-05-10
Add tool_deps tool to allow triggering installation of local tool dep…

innovate-invent:
…endencies

The contribution docs don't detail how to do tests. This tool requires that a tool has already been manually installed into galaxy.

It also isn't clear how to provide a command line wrapper to execute the python script

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -10,5 +10,6 @@ Commands
    commands/run-data-managers
    commands/setup-data-libraries
    commands/shed-tools
+   commands/tool-deps
    commands/workflow-install
    commands/workflow-to-tools
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -10,6 +10,6 @@ Commands
    commands/run-data-managers
    commands/setup-data-libraries
    commands/shed-tools
-   commands/tool-deps
+   commands/install-tool-deps
    commands/workflow-install
    commands/workflow-to-tools
--- a/docs/commands/install-tool-deps.rst
+++ b/docs/commands/install-tool-deps.rst
@@ -9,4 +9,4 @@ Usage
 .. argparse::
    :module: ephemeris.tool_deps
    :func: _parser
-   :prog: tool-deps
+   :prog: install-tool-deps
--- a/docs/commands/install-tool-deps.rst
+++ b/docs/commands/install-tool-deps.rst
@@ -1,4 +1,4 @@
-Workflow-install
+Install-tool-deps
 =================
 
 .. automodule :: ephemeris.tool_deps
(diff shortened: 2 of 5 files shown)


src/ephemeris/tool_deps.py
```diff
@@ -0,0 +1,51 @@
...
+
+from ephemeris import get_galaxy_connection
+from ephemeris.common_parser import get_common_args
+
+
+def _parser():
+    parent = get_common_args()
+    parser = argparse.ArgumentParser(parents=[parent])
+    parser.add_argument("-t", "--tool", help='Path to a tool file or tool_conf file')
+    parser.add_argument("-i", "--id", help='Comma seperated list of tool ids')
```
bgruening: You could use nargs here.

docs/commands.rst
```diff
@@ -10,5 +10,6 @@ Commands
    commands/run-data-managers
    commands/setup-data-libraries
    commands/shed-tools
+   commands/tool-deps
```
bgruening: Maybe install-tool-deps?

bgruening: If you could add sub-commands like show, install, remove, reinstall ... then this name would make sense imho. Do you plan to add those?

innovate-invent: I was thinking of having sub-commands but wanted feedback before implementing all of that :)

I'll take a crack at them tomorrow.

innovate-invent: Looks like bioblend provides no means of inspecting or uninstalling a tool dependency

mvdbeek:
This requires either the tool xml file or a tool_conf.xml, neither of which are formats we deal with in ephemeris and that are very local to Galaxy. I'd suggest a simple yaml list of tool ids.

innovate-invent:
@mvdbeek It does take a list of tool ids on the command line. The tool_conf.xml support is useful when deploying a container and ephemeris is connecting to localhost with direct access to galaxies filesystem.

mvdbeek:
I understand the utility, but in principle that's not the scope of ephemeris. Still this mostly works via the API, so if you can add loading a plain list of tool ids via `yaml.safe_load` and make that the default for `-t` I'm OK to merge this.

docs/commands/install-tool-deps.rst
```diff
@@ -0,0 +1,12 @@
+Workflow-install
```
mvdbeek:
```suggestion
Install-tool-deps
```

innovate-invent: Thanks for catching that

mvdbeek:
There are some linting errors now (https://travis-ci.org/galaxyproject/ephemeris/jobs/543810023#L475), the other failures should have been temporary and unrelated.

innovate-invent:
The current lint failure is not relevant to this project. The security implication is that it could fill RAM and crash the system if a malicious xml file is provided.

PR #135, opened 2019-06-04
Fix default message check

mvdbeek:
requests.exceptions.ConnectionError Exceptions don't have a body
attribute, but bioblend's ConnectionError `__str__` includes
the body:
```
    def __str__(self):
        return "{0}: {1}".format(self.args[0], self.body)
```


mvdbeek:
Thanks Enis!

PR #138, opened 2019-06-19
Data library creation would hang forever if you had a single non-ok command ever

erasche:
- supports legacy + non
- if legacy, fetches a list of all upload1 jobs and check states of those in a loop
- if normal, now checks the actual upload job that was scheduled.

### DIFF
--- a/src/ephemeris/setup_data_libraries.py
+++ b/src/ephemeris/setup_data_libraries.py
@@ -77,7 +77,7 @@ def create_legacy(gi, desc):
                         file_type=has_items['ext']
                     )
                 except Exception:
-                    log.error("Could not upload %s to %s/%s", has_items['url'], lib_id, base_folder_id)
+                    log.exception("Could not upload %s to %s/%s", has_items['url'], lib_id, base_folder_id)
         return None
 
     populate_items(folder_id, desc)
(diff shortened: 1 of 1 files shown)


src/ephemeris/setup_data_libraries.py
```diff
@@ -76,7 +76,7 @@ def populate_items(base_folder_id, has_items):
                         folder_id=base_folder_id,
                         file_type=has_items['ext']
                     )
-                except:
+                except Exception:
```
mvdbeek: Can you either use `log.exception` or `exc_info=True` for the logging? this may contain important information

erasche: Sure

erasche: It occurs when the filetype is unknown on the galaxy server, that's the only example I can think of.

erasche:
@mvdbeek is this ok?

src/ephemeris/setup_data_libraries.py
```diff
@@ -150,19 +154,32 @@ def normalize_items(has_items):
...
+            # Just have to check that all upload1 jobs are termianl.
+        else:
+            # Otherwise get back an actual list of jobs
+            for job in jobs:
+                if 'jobs' in job:
+                    for subjob in job['jobs']:
+                        job_ids.append(subjob['id'])
+
+        while True:
+            job_states = [jc.get_state(job) in ('ok', 'error') for job in job_ids]
```
mvdbeek:
```suggestion
            job_states = [jc.get_state(job) in ('ok', 'error', 'deleted', '') for job in job_ids]
```
that'd be all the terminal job states Galaxy knows about ... there's also paused which is sort-of terminal, but I guess that can't really happen in upload jobs.

erasche:
Thanks!

Issue #140, opened 2019-06-28
Workaround needed: Repos that were updated without a version bump are not installable to the new revision


rhpvorderman:
See #https://github.com/galaxyproject/galaxy/issues/8243

rhpvorderman:
I have taken a look at the "check if already installed code" and it is fairly inefficient. I can imagine it will also take quite long of you install thousands of tools. The checking algorithm is particularly slow. So that might need some work too. Unfortunately I have to leave office today at 11.30. But monday I will try and fix a few issues. Luckily we have plenty of testing in place so it should be possible to introduce a fix without introducing regressions.

rhpvorderman:
@erasche. Can you give me one or more tools+revisions that refuse to update? That will give me some examples to work from. I will also check our own galaxy instance.

erasche:
@mvdbeek reported in https://github.com/galaxyproject/galaxy/issues/8243#issuecomment-506479185 that iuc/fastqc tries repeatedly for version `e7b2202befea`.

I will have to look for others later., conference starts today.

nsoranzo:
Fixed by https://github.com/galaxyproject/galaxy/pull/8815

PR #141, opened 2019-07-01
Proper and fast check if repos are installed.

rhpvorderman:
Works around #140 
@erasche , can you run this on the EU galaxy tool lists and report your findings? (Maybe make an image of the server first... All the testing succeeds but... you never know.).

### DIFF
--- a/src/ephemeris/shed_tools.py
+++ b/src/ephemeris/shed_tools.py
@@ -52,7 +52,7 @@ from galaxy.tools.verify.interactor import GalaxyInteractorApi, verify_tool
 
 from . import get_galaxy_connection, load_yaml_file
 from .ephemeris_log import disable_external_library_logging, setup_global_logger
-from .get_tool_list_from_galaxy import GiToToolYaml, the_same_repository, tools_for_repository
+from .get_tool_list_from_galaxy import GiToToolYaml, tools_for_repository
 from .shed_tools_args import parser
 from .shed_tools_methods import complete_repo_information, flatten_repo_info, VALID_KEYS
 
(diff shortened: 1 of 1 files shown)


rhpvorderman:
I also tackled the very inefficient search mechanism. Now it makes a set of installed repos and compares against the set. Instead of looping over the list of installed tools.
This should create some speedup on a server with thousands of tools installed.

erasche:
Absolutely, I'll try to do that this evening! Thanks so much for tackling this

erasche:
We live dangerously so it's running in prod. https://build.galaxyproject.eu/job/usegalaxy-eu/job/install-tools/117/console Hopefully it works!

rhpvorderman:
It has to call the galaxy api quite a lot still. So if it is slow I can think of another way to speed it up... But that will require some more work. I hope this is sufficient.

rhpvorderman:
Hmm still running for an hour...
It's a bit annoying that it is not more verbose..

erasche:
This seems to have taken even longer, but not sure if that's something on our end? We can keep trying. The build took 12 hours and was killed(ish) at midnight by an automated process.

rhpvorderman:
@erasche. There was no way to check if this workaround was faster by putting it to a real production test. 
I guess the API call for each of the installed tools takes way too much time. This is the only way to determine if galaxy is actually going to skip the tool. But this check probably is just as slow as actually trying to install a new tool. I am affraid it is not really possible to workaround around the slowness of the API in that case.

I removed the api check. By using sets to determine if a tool is already installed this should speed up the process a bit, given the quite big number of tools on the galaxy server.

erasche:
> There was no way to check if this workaround was faster by putting it to a real production test.

Oh, all good. We're happy to be that test :)

rhpvorderman:
@erasche. 
I think this problem is not solvable unless the galaxy api is fixed.
Deducing whether galaxy is going to install a tool or not requires an API call for each tool on the galaxy server, which is extremely slow.
Alternatively the list of all available repos on the toolshed can be downloaded to make a list of installable revisions. If a galaxy tool is not an installable revision the API call method can be used. This **might** save some time, but I doubt it. It also introduces the overhead of having to download the repolists from each toolshed (~10 mb of data) for each installed yaml. Which is not worth it.

Nevertheless, comparing using sets should save some time on the install. So that change can be kept from this PR.

rhpvorderman:
@mvdbeek can you have a look at this PR? It does a set comparision to check if stuff is already installed. As such it should speedup the install scripts on big galaxies with multiple tools installed.

mvdbeek:
Well, this doesn't optimise the bottleneck and introduces more code. If we do want to optimise the code (I don't think we should) I'd just build sets of tuples up front and use the asymmetric difference.

rhpvorderman:
Close this as it does not solve the bottleneck as @mvdbeek stated.

PR #142, opened 2019-07-16
Release to PyPI on tag

mvdbeek:


nsoranzo:
I'd use a deploy stage, so it runs after all `test` builds are successful (and only once). Something like:
```yaml
matrix:
  ...
  include:
    - stage: deploy
      python: '3.5'
      if: tag IS present
      install:
        - python -m pip install twine
      script:
        - python setup.py sdist bdist_wheel
        - twine check dist/*
        - twine upload --skip-existing dist/*
```

PR #146, opened 2019-10-05
When polling for repo install status, ensure the correct revision is being checked

natefoo:
Otherwise any version of the repo that's already installed and encountered first in the loop will cause the poll to immediately return success.


natefoo:
Thanks @mvdbeek!

mvdbeek:
Sure, thanks a lot, looking forward to latest versions on test!

PR #149, opened 2019-10-23
setup_data_libraries: pass description when creating folders

abretaud:
A little fix for setup_data_libraries in legacy mode: folders' descriptions were not passed to galaxy
ping @bgruening


bgruening:
Awesome! Is that also possible for files or only folders?

mvdbeek:
Not sure, it's not exposed in bioblend: https://bioblend.readthedocs.io/en/latest/api_docs/galaxy/all.html#bioblend.galaxy.libraries.LibraryClient.upload_file_from_url

bgruening:
@mvdbeek does Galaxy expose the API. i.e. do we need to fix bioblend?

PR #154, opened 2020-02-24
Fix installing tool dependencies from yaml list

mvdbeek:


nsoranzo:
Can you make a release with this fix, in case we need it for the Admin Training next week?

mvdbeek:
Sure, let me revamp the timeout handling as well.

Issue #153, opened 2020-02-07
wait_for_install is too strict


mvdbeek:
Galaxy automatically picks the newest installable revision, so we can't simply wait for the requested revision to appear. We could instead pick up the revision on the repo that is in installing mode. Could be a problem is simultaneously installing multiple repos, but that is problematic in so many more ways that I think this wouldn't be a bad implementation.

https://github.com/galaxyproject/ephemeris/blob/master/src/ephemeris/shed_tools.py#L433
xref https://github.com/galaxyproject/usegalaxy-tools/pull/109

PR #158, opened 2020-04-19
Wait for the correct repository

mvdbeek:
This fixes https://github.com/galaxyproject/ephemeris/issues/153.

### DIFF
--- a/src/ephemeris/shed_tools.py
+++ b/src/ephemeris/shed_tools.py
@@ -437,30 +437,31 @@ class InstallRepositoryManager(object):
         name = repository['name']
         owner = repository['owner']
         changeset_revision = repository['changeset_revision']
-        installed_repo_list = self.tool_shed_client.get_repositories()
-        correct_owner_and_name_status = [r for r in installed_repo_list if r['name'] == name and r['owner'] == owner and r['status'] in NON_TERMINAL_REPOSITORY_STATES]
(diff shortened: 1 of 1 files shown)


src/ephemeris/shed_tools.py
```diff
@@ -423,20 +432,41 @@ def wait_for_install(self, repository, log=None, timeout=3600):
...
-                        else:
-                            time.sleep(10)
+                installed_repo = self.tool_shed_client.show_repository(installing_repo_id)
+                status = installing_repo['status']
+                if  status == 'Installed':
+                    return True
+                elif status == 'Error':
+                    return False
+                else:
+                    assert status in NON_TERMINAL_REPOSITORY_STATES, "Repository name '%s', owner '%s' in unknown status '%s'" % (name, owner, status)
```
rhpvorderman, line 468: Can this be:
```
elif status in NON_TERMINAL_REPOSITORY_STATES:
    time.sleep(10)
else:
    raise AssertionError(...)
```
?
There are some [concerns regarding the usage of assert](https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html). While I do not think this is necessarily applicable here, maybe it is better to use `raise` explicitly.  
Also the elif else construct is a little more explicit.

mvdbeek:
Awesome, thanks for the review!

PR #159, opened 2020-04-22
get all tools when searching for tool ids for testing (get_tools instead of get_tool_panel)

cat-bro:
When shed-tools looks for tests, it looks for tool ids within the output of get_tool_panel.  There is generally only one version of a tool in the panel but there could be several versions of the tool installed.  I'd like to be able to run tests for older revisions of tools.  Currently if the --revisions arg does not correspond to the version in the panel, no tests are run.

My proposed change is to look in the output of get_tools instead of get_tool_panel if the revision is provided.

I'm not sure what should happen when there is no changeset revision - I've defaulted using get_tool_panel in this case to look for what is usually the most recent version, but I could change this so that it runs tests from all versions.

Catherine, Galaxy Australia

### DIFF
--- a/src/ephemeris/get_tool_list_from_galaxy.py
+++ b/src/ephemeris/get_tool_list_from_galaxy.py
@@ -19,7 +19,12 @@ def get_tool_panel(gi):
     return tool_client.get_tool_panel()
 
 
-def tools_for_repository(gi, repository):
+def get_tools(gi):
+    tool_client = ToolClient(gi)
+    return tool_client.get_tools()
+
+
+def tools_for_repository(gi, repository, all_tools=False):
     tool_shed_url = repository.get('tool_shed_url')
     name = repository['name']
     owner = repository['owner']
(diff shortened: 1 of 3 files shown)


mvdbeek:
Thanks, this looks good, but could you you make this a non-default option please ?

cat-bro:
Hi @mvdbeek, thanks for reviewing.  What is meant by 'make this a non-default option'?  That instead of `tools_for_repository` having the extra argument `all_tools`, it should be calling `get_tools` instead of `get_tool_panel` as standard behaviour?

mvdbeek:
Yes, if you can add something like `--all_versions` in https://github.com/galaxyproject/ephemeris/blob/69c053742bf54201edef039c02299fa6ef1bd41b/src/ephemeris/shed_tools_args.py#L204

cat-bro:
Hi @mvdbeek, I've added this as boolean `--test_all_versions` (default False)

PR #164, opened 2020-10-28
Allow using client_test_config with shed_tools test.

jmchilton:
xref https://github.com/galaxyproject/galaxy/pull/10548


jmchilton:
Well crap, we have to drop Python 2.7 support to upgrade Galaxy components.

Issue #117, opened 2019-01-03
Tool testing improvements


mvdbeek:
These are a couple of things I found while implementing tool tests:

- tests only run of the test data is available from where the tests are triggered (this should be fixed with a galaxy-lib update containing https://github.com/galaxyproject/galaxy/pull/7160)
- tests seem to require an admin API key, that shouldn't be necessary with https://github.com/galaxyproject/galaxy/pull/7160
- We create a new history for every test, even if this test is for the same tool
- We just raise JobOutputExceptions, these need to be displayed properly
- We should have some reporting, ideally re-using some things from planemo

selten:
@mvdbeek would it also be an idea to add the ability to remove the histories after all tests are finished (or at least as a command line option)?

mvdbeek:
Yep, that would be a good option. I've added this along with some other things we've recently come across

selten:
`shed-tools test` should have an option to exit with exit code 1 in case of failed tests. For me it seems to exit with exit code 0 even if it has failed tests.

PR #169, opened 2021-02-13
Add shed-tools test options

selten:
*  Add option to cleanup the history after running the test
*  Add option to fail when an error is encounted for all shed-tools options.
Solves part of https://github.com/galaxyproject/ephemeris/issues/117

### DIFF
--- a/src/ephemeris/shed_tools.py
+++ b/src/ephemeris/shed_tools.py
@@ -37,6 +37,7 @@ import datetime as dt
 import json
 import os
 import re
+import sys
 import time
 from collections import namedtuple
 from concurrent.futures import thread, ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


src/ephemeris/shed_tools.py
```diff
@@ -627,10 +637,16 @@ def main():
...
             client_test_config_path=args.client_test_config,
+            cleanup_histories=args.cleanup_histories,
+            fail_on_error=args.fail_on_error,
         )
     else:
         raise NotImplementedError("This point in the code should not be reached. Please contact the developers.")
 
+    # Validate if all tools installed correctly
+    if args.fail_on_error and install_results and install_results.errored_repositories > 0:
+        log.info("The %s of a tool has failed. Exiting." % args.action)
```
rhpvorderman, line 648: log.error would be more appropriate.
Alternatively: since log + exit is usted. It may be better to use `raise`  an error instead.

rhpvorderman (changes requested):
It looks good to me. Can you add a test for this behavior?

PR #172, opened 2021-04-14
Simplify tox.ini, base testing on galaxy-compose

rhpvorderman:
This simplifies using tox with any version of python, which is convenient.

Also I made an attempt to get a 'quicker' galaxy. I made a [galaxy-compose](https://github.com/LUMC/docker-galaxy-compose) image which is basically a refactoring of the galaxy-k8s image for docker compose setups. Due to starting a postgres, nginx and galaxy container simultaneously the startup time is improved and testing can commence earlier. This shaves 50 seconds of the testing (7m20 ->6m30). 

The galaxy-compose image can easily be updated to newer versions of galaxy and is therefore quite suitable for testing. Images are build weekly.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,7 +24,5 @@ jobs:
           python -m pip install --upgrade pip
           pip install tox coveralls
       - name: Test with tox
-        run: |
-          TOXENV=$(echo $TOXENV | sed 's/\.//') tox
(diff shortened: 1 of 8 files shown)


rhpvorderman:
It turns out galaxy wait waits 30 seconds without checking anything. Eliminating that from the startup cuts the testing down to 5 minutes.

Issue #180, opened 2021-09-22
Global install_resolver_dependencies: false in yaml file does not work


mvdbeek:
https://github.com/galaxyproject/ephemeris/blob/c6aea18059bdbf405a900288a0ff9e5113c5393b/src/ephemeris/shed_tools_args.py#L123 with https://github.com/galaxyproject/ephemeris/blob/master/src/ephemeris/shed_tools.py#L609 means the command line argument default value takes precedence, which is True.

PR #181, opened 2021-09-23
Fix precedence of commandline options and yaml options

mvdbeek:
commandline options should take precedence over options specied in yaml
file, but default values should be used if options are not set in yaml
file. Fixes https://github.com/galaxyproject/ephemeris/issues/180


natefoo (approved):

martenson:
Just hit this again. What is more confusing is that the [docs](https://ephemeris.readthedocs.io/en/latest/commands/shed-tools.html#named-arguments) describe the default (`True`) as related to _skipping_. Which seems to be just a confusing error.

```
--skip-install-resolver-dependencies, --skip_install_resolver_dependencies
Skip installing tool dependencies through resolver (e.g. conda). Will be ignored on galaxy releases older than 16.07. Can be overwritten on a per-tool basis in the tools file

Default: True
```

martenson:
docs rebuild does not update it, our docs generation logic maybe cannot make sense of two args storing in the same var in this way

https://github.com/galaxyproject/ephemeris/blob/d60bea5485f89f592da77eb22b31b10a3eccc21e/src/ephemeris/shed_tools_args.py#L136C1-L152C10

PR #187, opened 2022-11-18
Set Library Permissions

mira-miracoli:
I added a tool to set the permissions of ALL datasets in a given data library (in folders and subfolders) at once to the given roles.
It is a workaround for one point of this [Issue](https://github.com/galaxyproject/galaxy/issues/15007), I created.

Usage:
```
usage: set_library_permissions [-h] [-v] [-g GALAXY] [-u USER] [-p PASSWORD] [-a API_KEY] [--library LIBRARY] [--roles ROLES]
```

Currently the default loglevel is INFO. I don't have experience with this so please tell me if I should change it.

### DIFF
--- a/src/ephemeris/set_library_permissions.py
+++ b/src/ephemeris/set_library_permissions.py
@@ -121,10 +121,8 @@ def main():
         )
     set_permissions(gi, library_id=args.library, role_ids=args.roles, auto=args.yes)
     log.info(
-        "\nThis command script uses bioblend galaxyAPI to set ALL permissions of ALL datasets"
-    )
-    log.info(
-        "in given library to given roles. Be careful and cancel with Crtl+C if unsure.\n"
+        "\nThis script uses bioblend to update ALL permissions of ALL datasets in a"
+        "specified library to the given roles. Be careful and cancel if unsure\n"
     )
 
 
--- a/src/ephemeris/set_library_permissions.py
+++ b/src/ephemeris/set_library_permissions.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 """Tool to set permissions for all datasets of a given Galaxy Data Library"""
 
-import sys
 import argparse
+import sys
 import logging as log
 
 from bioblend import galaxy
--- a/src/ephemeris/set_library_permissions.py
+++ b/src/ephemeris/set_library_permissions.py
@@ -5,29 +5,10 @@ import argparse
 import logging as log
 from bioblend import galaxy
 import sys, time, os
-
+from rich.progress import Progress
 from .common_parser import get_common_args
 # Print iterations progress
-def printProgressBar (iteration, total, prefix = '', suffix = '', decimals = 1, length = 100, fill = '█', printEnd = "\r"):
-    """
-    Call in a loop to create terminal progress bar
-    @params:
-        iteration   - Required  : current iteration (Int)
-        total       - Required  : total iterations (Int)
-        prefix      - Optional  : prefix string (Str)
-        suffix      - Optional  : suffix string (Str)
-        decimals    - Optional  : positive number of decimals in percent complete (Int)
-        length      - Optional  : character length of bar (Int)
-        fill        - Optional  : bar fill character (Str)
-        printEnd    - Optional  : end character (e.g. "\r", "\r\n") (Str)
-    """
-    percent = ("{0:." + str(decimals) + "f}").format(100 * (iteration / float(total)))
-    filledLength = int(length * iteration // total)
-    bar = fill * filledLength + '-' * (length - filledLength)
-    print(f'\r{prefix} |{bar}| {percent}% {suffix}', end = printEnd)
-    # Print New Line on Complete
-    if iteration == total: 
-        print()
+
 
 def get_datasets(gi, library_id) -> [str]:
     objects = gi.libraries.show_dataset(library_id=library_id, dataset_id='')
@@ -49,17 +30,20 @@ def set_permissions(gi, library_id, role_ids):
     # Give User time to abort
     log.info('\nSuccess! %d datasets found. Processing can take up to %f min', total, est)
     t = 5
-    while t:
-        log.info('Starting in %d s ... Press Crtl+C to abort.', t)
-        time.sleep(1)
-        t -= 1
+    with Progress() as progress:
+        task_countdown = progress.add_task("[red]Starting in 5 seconds...", total=total)
+        t = 5
+        while t:
+            time.sleep(1)
+            t -= 1
+            progress.update(task_countdown, advance=1)
     # Process datasets
-    for item in datasets:
-        current = datasets.index(item) + 1
-        log.debug('Processing dataset %d of %d, ID=%s', current, total, item)
-        rows, columns = os.popen('stty size', 'r').read().split()
-        printProgressBar(iteration=current, total=total, length=(int(columns) - 3))
-        gi.libraries.set_dataset_permissions(dataset_id=item, access_in=role_ids, modify_in=role_ids, manage_in=role_ids)  
+        task = progress.add_task("[green]Processing datasets...", total=total)
+        for current in range(total):
+            log.debug('Processing dataset %d of %d, ID=%s', current, total, datasets[current])
+            rows, columns = os.popen('stty size', 'r').read().split()
+            gi.libraries.set_dataset_permissions(dataset_id=datasets[current], access_in=role_ids, modify_in=role_ids, manage_in=role_ids)  
+            progress.update(task, advance=1)
 
 def _parser():
     '''Constructs the parser object'''
--- a/src/ephemeris/set_library_permissions.py
+++ b/src/ephemeris/set_library_permissions.py
@@ -1,0 +1,100 @@
+#!/usr/bin/env python
+'''Tool to set permissions for all datasets of a given Galaxy Data Library'''
+
+import argparse
+import logging as log
+from bioblend import galaxy
+import sys, time, os
+
+from .common_parser import get_common_args
+# Print iterations progress
+def printProgressBar (iteration, total, prefix = '', suffix = '', decimals = 1, length = 100, fill = '█', printEnd = "\r"):
+    """
+    Call in a loop to create terminal progress bar
+    @params:
+        iteration   - Required  : current iteration (Int)
+        total       - Required  : total iterations (Int)
+        prefix      - Optional  : prefix string (Str)
+        suffix      - Optional  : suffix string (Str)
+        decimals    - Optional  : positive number of decimals in percent complete (Int)
+        length      - Optional  : character length of bar (Int)
+        fill        - Optional  : bar fill character (Str)
+        printEnd    - Optional  : end character (e.g. "\r", "\r\n") (Str)
+    """
+    percent = ("{0:." + str(decimals) + "f}").format(100 * (iteration / float(total)))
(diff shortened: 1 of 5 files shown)


src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,100 @@
...
+           datasets.append(objects[index]['id'])
+    if datasets == []:
+        sys.exit("No datasets in library!")
+    else:
+        return datasets
+
+def set_permissions(gi, library_id, role_ids):
+    log.info("Your library_id is " + library_id + "\n")
+    role_msg = "Your roles are: " + " ".join(role_ids)
+    log.info(role_msg)
```
hexylena, line 46: this way the string will only get properly templated *if* you're above the correct log level.
```suggestion
    log.info("Your roles are: %s", " ".join(role_ids))
```

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,100 @@
...
+    datasets = get_datasets(gi, library_id)
+    total = len(datasets)
+    est = total*3/60
+    # Give User time to abort
+    log.info('\nSuccess! %d datasets found. Processing can take up to %f min', total, est)
+    t = 5
+    while t:
+        log.info('Starting in %d s ... Press Crtl+C to abort.', t)
+        time.sleep(1)
+        t -= 1
```
hexylena, line 56:
```suggestion
```

I don't know if this is necessary? If you want to do this, we should default to "dry run" and then have a `--commit` flag that would actually do the change permissions call

hexylena (commented):
Some small review comments, this will be really nice to have :)

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,100 @@
...
+'''Tool to set permissions for all datasets of a given Galaxy Data Library'''
+
+import argparse
+import logging as log
+from bioblend import galaxy
+import sys, time, os
+
+from .common_parser import get_common_args
+# Print iterations progress
+def printProgressBar (iteration, total, prefix = '', suffix = '', decimals = 1, length = 100, fill = '█', printEnd = "\r"):
```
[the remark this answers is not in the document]
mira-miracoli: Sound cool on one hand, but would increase dependencies on the other, maybe when someone else starts using it here?

src/ephemeris/set_library_permissions.py
mira-miracoli: I have this because the API is quite slow in this case and it took ~50 min to change 940 datasets, so I would like to inform the user about the estimated time and give the option to abort before it's starting

hexylena: Yes, hence doing the converse:

- doing a dry run by default (which can give a time estimate0
- and then running if they request it

Rather than having a short window in which, if they don't read fast enough, things can start happening.

The other "safe" alternative is doing an `input()` which will block on user input (so you can't accidentally have it run by not reading fast enough about what's happening). And checking if they say `y` to continue.

src/ephemeris/set_library_permissions.py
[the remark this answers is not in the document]
natefoo: I used [rich](https://github.com/Textualize/rich) in [galaxy-upload](https://github.com/galaxyproject/galaxy-upload) but didn't have to worry about adding a dependency in that case since it was already a dependency of one of the Galaxy libs.

hexylena (commented):
looks great with the new progress bar, fantastic!

src/ephemeris/set_library_permissions.py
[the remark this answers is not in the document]
mira-miracoli: image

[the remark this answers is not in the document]
mira-miracoli: Just added that with the last commit :)

src/ephemeris/set_library_permissions.py
mira-miracoli: I added the y/n question

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,79 @@
...
+
+def set_permissions(gi, library_id, role_ids):
+    log.info("Your library_id is " + library_id + "\n")
+    log.info("Your roles are: %s", " ".join(role_ids))
+    datasets = get_datasets(gi, library_id)
+    total = len(datasets)
+    est = total*3/60
+    # Give User time to abort
+    log.info('\nSuccess! %d datasets found. Processing can take up to %f min', total, est)
+    if input("Do you want to continue? (y/n) ") == "y":
```
mvdbeek, line 32: A lot of ephemeris is meant to be used as a library, it's unfortunate to have `input` in that case. Can you make this optional ?

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,79 @@
...
+        with Progress() as progress:
```
mvdbeek, line 33: Is there a quiet/silent option, and can you use it ?

src/ephemeris/set_library_permissions.py
mvdbeek: > I have this because the API is quite slow in this case and it took ~50 min to change 940 datasets, so I would like to inform the user about the estimated time and give the option to abort before it's starting

Can you create an issue in Galaxy for this ? There's no intrinsic reason that this needs to be so slow.

src/ephemeris/set_library_permissions.py
mira-miracoli: I could add a -y flag, would that be okay?

src/ephemeris/set_library_permissions.py
mira-miracoli: Maybe loglevel = error could solve this?
I'll make a few small changes for that

src/ephemeris/set_library_permissions.py
mvdbeek: if the function's default argument is to be quiet, yes.

src/ephemeris/set_library_permissions.py
```diff
@@ -28,19 +28,21 @@ def set_permissions(gi, library_id, role_ids):
     total = len(datasets)
     est = total*3/60
     # Give User time to abort
-    log.info('\nSuccess! %d datasets found. Processing can take up to %f min', total, est)
-    if input("Do you want to continue? (y/n) ") == "y":
-        with Progress() as progress:
-            task = progress.add_task("[green]Processing datasets...", total=total)
+    log.info('\nSuccess! %d datasets found. Processing can take up to %0.02f min\n', total, est)
+    if auto:
```
mvdbeek, line 32: I think you're missing the `auto` argument in the function signature

mira-miracoli: yes, sorry

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,132 @@
...
+    else:
+        sys.exit(
+            "Specify library ID (--library myLibraryID) and (list of) role(s) (--roles roleId1,roleId2)"
+        )
+    set_permissions(gi, library_id=args.library, role_ids=args.roles, auto=args.yes)
+    log.info(
+        "\nThis command script uses bioblend galaxyAPI to set ALL permissions of ALL datasets"
+    )
+    log.info(
+        "in given library to given roles. Be careful and cancel with Crtl+C if unsure.\n"
```
hexylena, line 127:
```suggestion
        "\nThis script uses bioblend to update ALL permissions of ALL datasets in a"
        "specified library to the given roles. Be careful and cancel if unsure\n"
```

these could be merged

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def _parser():
+    """Constructs the parser object"""
+    parent = get_common_args()
+    parser = argparse.ArgumentParser(
+        parents=[parent], description="Populate the Galaxy data library with data."
+    )
+    parser.add_argument("--library", help="Specify the data library ID")
+    parser.add_argument("--roles", help="Specify a list of comma separated role IDs")
```
hexylena, line 78:
```suggestion
    parser.add_argument("--roles", nargs="+", help="Specify a list of comma separated role IDs")
```

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,132 @@
...
+            log.info("Operation cancelled by user. No changes were applied.\n")
+
+
+def _parser():
+    """Constructs the parser object"""
+    parent = get_common_args()
+    parser = argparse.ArgumentParser(
+        parents=[parent], description="Populate the Galaxy data library with data."
+    )
+    parser.add_argument("--library", help="Specify the data library ID")
```
hexylena, line 77:
```suggestion
    parser.add_argument("library", help="Specify the data library ID")
```

since this is mandatory, maybe makes sense to just require it outright

src/ephemeris/set_library_permissions.py
```diff
@@ -0,0 +1,132 @@
...
+            datasets.append(objects[index]["id"])
+    if datasets == []:
+        sys.exit("No datasets in library!")
+    else:
+        return datasets
+
+
+def set_permissions(gi, library_id, role_ids, auto):
+    log.info("Your library_id is " + library_id + "\n")
+    log.info("Your roles are: %s", " ".join(role_ids))
```
hexylena, line 28:
```suggestion
    log.info("Your library_id is %s", library_id)
    log.info("Your roles are: %s", " ".join(role_ids))
```

hexylena (approved):
looks good to me! great work @mira-miracoli

mira-miracoli:
Thank you for your reviews and suggestions @hexylena :)

hexylena:
Sorry @mira-miracoli, the linter has some additional requests for you to re-order the imports and add whitespace here and there to make it happy.

hexylena:
```
Traceback (most recent call last):
  File "/home/runner/work/ephemeris/ephemeris/.tox/py36-integration/bin/shed-tools", line 8, in <module>
    sys.exit(main())
 ...
  File "/home/runner/work/ephemeris/ephemeris/.tox/py36-integration/lib/python3.6/site-packages/bioblend/galaxyclient.py", line 266, in key
    headers = self.json_headers.copy()
AttributeError: 'GalaxyInstance' object has no attribute 'json_headers'
```

that looks unrelated, but not sure what's causing that

mira-miracoli:
Shall I change something or can we merge it?

hexylena:
@nsoranzo any ideas about this error?

nsoranzo:
That error was fixed in commit https://github.com/galaxyproject/bioblend/commit/22f2d8919b2144f45e1ff83df1431c11de8a92ed , which was first released in BioBlend v0.17.0 . That release also dropped support for Python 3.6 (EOL for almost a year now), which is the version used in the tests.
So my suggestion would be to update the GitHub workflows to use Python 3.7 .

nsoranzo:
You may also want to add a `python_requires` (e.g. https://github.com/galaxyproject/gravity/blob/main/setup.py#L39 ) which is metadata used by PyPI and pip.

natefoo:
> Someone with more privileges than me (@mvdbeek?) will need to change the branch protection requirements to now depend on the update test names since their name includes the python version.

Got it

mvdbeek:
Thanks @mira-miracoli and reviewers, very cool work!

PR #191, opened 2022-12-14
Prefer dashes instead of underscores in flags

natefoo:
This converts all ephemeris command flags that use underscores to dashes instead. Underscores continue to be supported but are hidden from the help text.

Underscores in UNIX command line flags are pretty non-standard, and are way more painful to type, to boot.

Also, entry points for `install-tool-deps` and `set-library-permissions` are installed in addition to their original underscore versions. The underscore versions of these two commands should be considered deprecated maybe be removed at some point, but the underscore flags can be kept indefinitely with little harm.

### DIFF
--- a/src/ephemeris/common_parser.py
+++ b/src/ephemeris/common_parser.py
@@ -3,6 +3,21 @@
 import argparse
 
 
+class HideUnderscoresHelpFormatter(argparse.HelpFormatter):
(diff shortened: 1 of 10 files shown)


mvdbeek:
It's pretty conflicted now, but happy to merge if you can resolve these

natefoo:
Done

PR #194, opened 2023-03-28
Add data_manager_mode argument fort IDC use

mvdbeek:
Idea is that we build bundles against the existing servers, then just pull the bundles into an instance that can write to cvmfs.


mvdbeek:
Merged as part of https://github.com/galaxyproject/ephemeris/pull/201

PR #195, opened 2023-04-04
Standardize CLI commands on - instead of _

hexylena:
We have one new command waiting for a release (`set-library-permissions`) that was documented as using `-` in the CLI examples, and then `install-tool-deps` which used `_`, and now gets an alias for `-`


hexylena:
@mvdbeek after this can I trouble you to mint a new release of ephemeris? I'd like to mention this utility during the data-libraries training

natefoo (approved):

natefoo:
This goes nicely with https://github.com/galaxyproject/ephemeris/pull/191 :sweat_smile:

PR #200, opened 2023-06-29
Script to split genomes.yml for IDC

jmchilton:


src/ephemeris/_split_data_manager_genomes.py
```diff
@@ -0,0 +1,202 @@
...
+    data_manager = data_managers[indexer]
+    assert data_manager, f"Could not find a target data manager for indexer name {indexer}"
+    return data_manager.tool_id
+
+
+class RunDataManager(BaseModel):
+    id: str
+    items: Optional[List[Any]] = None
+    params: Optional[List[Any]] = None
+    data_table_reload: Optional[List[str]] = None
```
mvdbeek, line 59: This shouldn't be needed. It just works on normal instances, and if it is data on cvmfs you first have to import the bundles this depends on, so the reload also doesn't do anything there.

jmchilton: I'm trying to model the inputs these tools consume... if we don't want to use the field that is fine I can drop the dependencies setup with this at line at line 122 but I think keeping the model correct is the right thing to do. Unless we update the samples and run_data_managers.py and such.

jmchilton: I would like to migrate all the models out and reuse them in ephemeris and I would like to setup models for genomes.yml and tools.yaml and such... there is some broader context here. I hope you're correct about the reloads not being needed though - ideally those would just happen (see my comment in IDC gitter)

jmchilton:
Given multiple people are hacking on this branch - a good point to merge? Everything is green.

mvdbeek:
Merged as part of https://github.com/galaxyproject/ephemeris/pull/201

PR #201, opened 2023-07-01
Enhancements to the IDC scripts

jmchilton:
- More unit tests.
- Add option of filtering by stage (and other things) in split script.
- More validation of generated artifacts (in particular tools YAML).
- More checks in the IDC lint (verify all reference indexers in genomes.yml are keyed in data_managers.yml and all data manager tools are from the main tool shed).


mvdbeek (approved):
green!

PR #208, opened 2023-11-29
Fix CI tests

mvdbeek:


tests/conftest.py
```diff
@@ -1,65 +1,58 @@
...
+                "environments": {
+                    "local_docker": {"runner": "local", "docker_enabled": True},
+                },
+            },
+        },
+        "admin_users": "test@bx.psu.edu",
+        "bootstrap_admin_api_key": "123456789",
+        "conda_auto_init": False,
+        "config_dir": str(config_dir),
+        "database_connection": f"sqlite:///{database}?isolation_level=IMMEDIATE",
```
jmchilton, line 37: The whole config as one JSON blob is randomly very pretty and very rewarding.

PR #202, opened 2023-07-02
Try to update test instance

jmchilton:

PR #211, opened 2024-01-31
Refactor build tracking

mvdbeek:
Just picking the non-debug commits from https://github.com/galaxyproject/ephemeris/pull/202


mvdbeek:
I guess we'll have to restore the cvmfs import check before merging this one.

PR #222, opened 2025-03-05
remove the duplicated argument for installing resolver deps

martenson:
this is confusing and likely
breaking docs generation showing a wrong default

xref https://github.com/galaxyproject/ephemeris/pull/181#issuecomment-2700723590


martenson:
with this pr:

Screenshot 2025-03-05 at 13 46 35

PR #225, opened 2025-03-17
Use verbose argument more often and add exit code to `shed-tools`

bernt-matthias:


src/ephemeris/shed_tools.py
```diff
@@ -733,6 +735,10 @@ def main(argv=None):
                 client_test_config_path=args.client_test_config,
             )
 
+    if install_results:
+        if len(install_results.errored_repositories) > 0:
+            sys.exit(1)
```
mvdbeek: Is it clear to the user why the exit code is non-zero ? If you do `sys.exit("Failing because error")` or something like that, that is also exit code 1

mvdbeek (approved):

src/ephemeris/shed_tools.py
```diff
@@ -733,6 +735,13 @@ def main(argv=None):
                 client_test_config_path=args.client_test_config,
             )
 
+    if install_results:
+        if len(install_results.errored_repositories) > 0:
+            log.error("There were errors for some repositories")
```
mvdbeek: Standard python would be raising an exception with that message. You can catch and print it in the `__main__` block

src/ephemeris/shed_tools.py
```diff
@@ -733,6 +739,13 @@ def main(argv=None):
                 client_test_config_path=args.client_test_config,
             )
 
+    if install_results and len(install_results.errored_repositories) > 0:
+        raise ToolInstallationException("There were errors for some repositories")
```
mvdbeek: Would it be more useful to say which repositories ?

mvdbeek (approved):

PR #234, opened 2026-05-11
Add script to install or tabulate (singularity) containers

bernt-matthias:
use cases:

1. run on the Galaxy machine and trigger caching of (singularity) containers (I'm using this since a long time on my machine: https://github.com/bernt-matthias/ufz-galaxy-scripts/blob/main/container/install_container.py). We could change this to be able to run also from remote machines (currently I'm checking if the path exists - since the resolvers also give a path if the container is not yet cached).
3. determine a table mapping docker URIs and container image names (paths). this table can then be used on machines that do not have access to CVMFS to download container images.

what the tool does

queries the list of tools from an instance (optionaly filter by regexes on the tool_ids and latest version)

for each tool:

- resolve the container (for cached singularity containers with a configured cache directory this will yield a path)
- if `--install_container` the resolver(s) will be called again with the `install` parameter set to `True` (if the determined path does not exist)
- if `--tabulate N` is used the `N`-th configured resolver is called and the result is printed together with the tool_id and the path that was determined in the 1st step (tab separated). The idea is that the instance configures a singularity container resolver (at index N) without a cache directory defined. This resolver will yield the original container URI (docker://...). If this resolver is defined after the cached / non-cached singularity container resolvers it should never be called in production.

### DIFF
--- a/src/ephemeris/install_container.py
+++ b/src/ephemeris/install_container.py
@@ -1,8 +1,27 @@
 """
-install (i.e. cache) containers for a set of tools
+Install or tabulate (singularity) containers
 
-tools can be selected by basic string matching (--filter)
-or version (--latest)
+Use cases:
+
+1. run on the Galaxy machine itself and trigger caching of (singularity) containers
+2. determine a table that maps docker URIs and cached container image paths.
+   this table can then be used on machines that do not have access to CVMFS
+   to download container images.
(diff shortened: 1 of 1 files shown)


bgruening:
Cool, that will be very useful. Can you please also add something to the docs.

PR #237, opened 2026-06-12
Add shed-tools validate subcommand

### LINKED ISSUE #236: Add a way to validate a tools.yaml against a Tool Shed without installing
_Posted by Claude (AI assistant) on behalf of @jmchilton._

## Problem

There's currently no way to check that a tool-install YAML (a `tools.yaml` / `*.yml.lock`) is correct without pointing `shed-tools` at a live Galaxy and actually installing. `shed-tools` exposes only `install`, `update`, and `test`, and all three require a `--galaxy` connection (`src/ephemeris/shed_tools_args.py`).

For curation repos like [`usegalaxy-tools`](https://github.com/galaxyproject/usegalaxy-tools) — where the files are large, fully revision-pinned `.yml.lock`s — it'd be valuable to validate in CI/pre-commit that:

1. the YAML is structurally valid (required keys present, right shape), and
2. every referenced repo (`name`/`owner`) and **every pinned `changeset_revision`** actually exists/installable on the target Tool Shed.

## Current gap

Beyond there being no `validate` subcommand, explicitly-pinned revisions are *never* checked against the Tool Shed — not even during a real install. `get_changeset_revisions()` only queries the shed when resolving "latest":

```python
# src/ephemeris/shed_tools_methods.py
if repository.get("changeset_revision") is None or force_latest_revision:
    ts = ToolShedInstance(url=repository["tool_shed_url"])
    installable_revisions = ts.repositories.get_ordered_installable_revisions(name, owner)
    if not installable_revisions:
        raise LookupError(f"Repo does not exist in tool shed: {repository}")
    repository["changeset_revision"] = installable_revisions[-1]
```

When a `changeset_revision` is already set (the lock-file case), this is skipped entirely and the hash is passed straight to Galaxy's install API. A typo'd or no-longer-installable revision only surfaces as a Galaxy-side install failure.

## Proposed solution

A no-install validation path, e.g. `shed-tools validate <tools.yaml> [--tool-shed URL]` (or a `--check-only` flag on `install`), that requires **no** `--galaxy`. The building blocks already exist:

- **Structural validation** — `src/ephemeris/_config_models.py` already has `read_tools()` / the `RepositoryInstallTargets` Pydantic model, but the install path uses plain `load_yaml_file` and never invokes it. Reuse `read_tools()` here.
- **Shed-side validation** — `ToolShedInstance(...).repositories.get_ordered_installable_revisions(name, owner)` already talks only to the Tool Shed (no Galaxy). Loosen the guard above into a standalone check: assert the repo exists and that every listed `revision` is in `installable_revisions`.

### Two concrete things it would catch that nothing catches today
- Malformed YAML / missing required keys, before any deploy.
- Pinned revisions that no longer resolve on the shed (the entire content of a `.yml.lock`).

### Open questions
- New `validate` subcommand vs `--check-only`/`--dry-run` on `install`?
- Should structural validation be unified on the Pydantic model so the install path validates too (vs `usegalaxy-tools`' separate pykwalify `.schema.yml`)?
- Validate latest-resolvable only, or assert exact pinned-revision membership (stricter — recommended)?

jmchilton:
> Drafted by Claude (AI assistant) on behalf of @jmchilton.

## What

Adds a `shed-tools validate` subcommand: no-install, no-Galaxy validation of a tools file / `.yml.lock` against a Tool Shed. Closes #236.

- **Structural check** reuses the `RepositoryInstallTargets` pydantic model via `read_tools` (same parsing as install).
- **Shed check** asserts each repository exists and every pinned revision is installable, querying each repo once.
- `--structural-only` skips all network access — suitable for offline pre-commit / CI on revision-pinned lock files.
- The subcommand needs no Galaxy connection; it deliberately does not inherit the Galaxy connection args.

```
shed-tools validate tools.yml                 # structure + shed
shed-tools validate tools.yml --structural-only   # offline, no network
```

## Hardening (second commit)

- `extra=forbid` on `RepositoryInstallTarget(s)` so a typo'd key (e.g. `revision:` for `revisions:`) fails structurally instead of silently passing with the pin unchecked — the failure mode the validator exists to catch. Safe: the install path reads raw YAML and does not use the model; the only model consumers are `validate`, the IDC generator (known keys only), and tests.
- `read_tools` rejects a non-mapping / empty YAML root with a clear error instead of a `TypeError` from `**`-splatting `None`.

## Tests

`tests/test_shed_tools_validate.py` fakes the Tool Shed via monkeypatch (no network): structural-only good/bad, full good/bad-revision, missing file, no input, unknown-key rejection, empty/non-mapping root, query-once-per-repo. mypy clean.

## Comment cleanup (rolled in from #238)

Per @mvdbeek's review on #238, the first commit tidies two comments that landed with that merge: drops the run-data-managers comment wrongly claiming Galaxy 26.0 requires a history to execute tools (the real cause was a session without an associated history), and trims the verbose `galaxy-app` pin comment in `dev-requirements.txt`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,14 @@ History
 .. to_doc
 
 ---------------------
+0.10.12.dev0
+---------------------
+
+* Add ``shed-tools validate`` to check a tools file structurally and against a
+  Tool Shed (every pinned ``changeset_revision``) without installing into Galaxy.
+  `Issue 236`_
+
+---------------------
 0.10.11 (2025-09-26)
 ---------------------
 
@@ -282,6 +290,7 @@ History
   and adapt them for usage as a library.
 
 .. github_links
+.. _Issue 236: https://github.com/galaxyproject/ephemeris/issues/236
 .. _Pull Request 222: https://github.com/galaxyproject/ephemeris/pull/222
 .. _Pull Request 220: https://github.com/galaxyproject/ephemeris/pull/220
 .. _Pull Request 212: https://github.com/galaxyproject/ephemeris/pull/212
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,10 +4,8 @@ coverage
 pytest
 pytest-cov
 galaxy-test-driver
-# Pin galaxy-app to the same release line as the rest of the Galaxy stack.
-# It is only pulled in transitively (via galaxy-test-driver), so pip is free to
-# backtrack it to an older release than galaxy-data/galaxy-util, producing a
-# skewed install whose galaxy.model import fails during test collection.
+# Pulled in only transitively via galaxy-test-driver; pin it to keep the Galaxy
+# stack on one release line (otherwise pip can backtrack it, skewing the install).
 galaxy-app>=26
 
 #Building Docs
--- a/src/ephemeris/_config_models.py
+++ b/src/ephemeris/_config_models.py
@@ -10,7 +10,7 @@ from pydantic import (
 StrOrPath = Path | str
 
 
-class RepositoryInstallTarget(BaseModel):
+class RepositoryInstallTarget(BaseModel, extra=Extra.forbid):
     name: str
     owner: str
     tool_shed_url: str | None = None
@@ -22,7 +22,7 @@ class RepositoryInstallTarget(BaseModel):
     install_resolver_dependencies: bool | None = None
 
 
-class RepositoryInstallTargets(BaseModel):
+class RepositoryInstallTargets(BaseModel, extra=Extra.forbid):
     """ """
 
     api_key: str | None = None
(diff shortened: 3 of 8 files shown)


bgruening:
@arash77 this looks interesting. Can you please point John to the scripts that you have been cooking - removing the non-installable revisions etc? Maybe that is interesting for whatever @jmchilton is cooking here.

arash77:
Sure! You can find the scripts for filtering out the non-installable tool revisions in this PR: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/pull/929

arash77:
Should we also allow/validate the top-level  `install_tool_dependencies` ,  `install_repository_dependencies` , and  `install_resolver_dependencies`  keys?
Currently,  `shed-tools validate`  fails on files that define these defaults because  `RepositoryInstallTargets`  uses  `extra=Extra.forbid`  but doesn't declare them.

PR #213, opened 2024-02-01
Dm parameters

mvdbeek:

PR #239, opened 2026-08-11
DM parameters updated

lldelisle:
Hi,
This is #213 but updated to fit the new code.

I think this feature is really needed.

Warning, this adds a new dependency (types-boltons to use 'remap' introduced by @mvdbeek in #213).


lldelisle:
@natefoo is it close to what you are using for BRC+ DM automation scripts?

mvdbeek:
Fo the IDC there is https://github.com/galaxyproject/idc/pull/89, which generates workflows directly, so no need to parameterize via the tools API for this purpose. There’s no problem with merging this if you need it locally but I think we should test out the IDC PR and pull the reusable parts into ephemeris.

lldelisle:
Hi,
I don't need to merge this. I am just trying to have a repo for my university instance with a 'idc-like' repo so users can request local genomes and everything is more or less automatic. But I will check what I can reuse from the IDC PR. Thanks

